PR #5, opened 2026-09-04
fix(scanner): re-enrich the local node on the peer refresh tick

### LINKED ISSUE #4: Local node's GPU/CPU/memory enrichment is never refreshed after startup
The discovery directory's enrichment for the **local** node is captured once and then frozen. Peers converge; self does not.

## What I measured

Ubuntu node with an RTX 3060, paired with a macOS node, v0.1.1-463 from the `.deb`.

I loaded a 7 GB model into LM Studio with full GPU offload. `nvidia-smi` went from 121 MiB to 9329 MiB. `node-info` served the new figure within its usual cadence — `curl 127.0.0.1:14318/v1/node-info` returned `vram_used_bytes: 9782165504` with `telemetryValid: true, msSince: 280`. The desktop UI's VRAM graph showed the step too, because it reads the node directly.

The discovery snapshot did not move. `discovery:get-nodes` kept returning `vram_used_bytes: 126877696` — the value from daemon start — for as long as I watched, checked at 20 s intervals over several minutes. Restarting the broker made it correct again, until the next change.

## Why

`onBrowse` deliberately skips our own entry (`daemon.go`: *"Never let a browse event touch our own entry. Self is registry-driven"*), so the path that re-enriches every peer on every browse never runs for self. `publishSelf` — the only thing that enriches self — is called at startup, on service (un)register, and after an identity/address change. `refreshPeersLoop` refreshes advertised addresses, the mesh, cluster identity and models on its 15 s tick, but nothing re-reads self's node-info.

Worth noting that the comment above `noteNodeInfo` already states the intended behaviour:

> The enrichment sweep repeats every peerRefreshInterval for as long as a node is advertised

That holds for peers, not for self.

## Who sees it

Only consumers of the discovery snapshot — which includes `nvpair-job-scheduler`, since the broker fans the same snapshot to it. The UI is unaffected, which is why this is easy to miss.

## Repro

1. Pair two nodes; note the local node's `gpus[].vram_used_bytes` via `discovery:get-nodes`.
2. Load a large model with GPU offload on that node.
3. `curl 127.0.0.1:14318/v1/node-info` — fresh.
4. `discovery:get-nodes` — unchanged, indefinitely.

## A fix that works

Calling `d.publishSelf()` on the `refreshPeersLoop` tick converges the directory within 20 s. Verified live in both directions, loading and unloading. It does mean one self `node-updated` per tick, since `publishSelf` emits unconditionally; the tidier alternative is an `applyInfo` path with a `changed` check, mirroring `applyModels`. Happy to open a PR for either shape if useful.

canja006:
Fixes #4.

`onBrowse` deliberately skips our own entry, so the path that re-enriches every peer on every browse never runs for self. `publishSelf` — the only thing that enriches self — is called at startup, on service (un)register, and after an identity or address change. Nothing was left to converge the local node's GPU, CPU and memory figures, so they stayed at whatever they were when the daemon started.

This calls `publishSelf()` on the `refreshPeersLoop` tick that already refreshes advertised addresses, the mesh, cluster identity and models.

**Verified on hardware**, two paired nodes (Ubuntu + RTX 3060, macOS):

- load a model with GPU offload → `discovery:get-nodes` follows within 20 s, where before it never moved at all
- unload → follows back
- `node-info` and the desktop UI were correct throughout; only the discovery snapshot was frozen

`go vet` and `go test ./...` clean in `services/nvpair-node-scanner`.

One tradeoff worth flagging: `publishSelf` emits `node-updated` unconditionally, so this adds one self update per 15 s tick. The tidier alternative is an `applyInfo` path with a `changed` check, mirroring `applyModels` — happy to switch to that shape if you prefer it.

### DIFF
--- a/services/nvpair-node-scanner/daemon.go
+++ b/services/nvpair-node-scanner/daemon.go
@@ -1467,6 +1467,17 @@ func (d *daemon) refreshPeersLoop(ctx context.Context) {
 			}
 			d.refreshClusterIdentityOnce()
 			d.refreshModelsOnce()
+			// Re-enrich self. onBrowse deliberately skips our own entry (self
+			// is registry-driven), and publishSelf is otherwise only called
+			// when the registry moves -- a service (un)register, an identity
+			// or address change. Nothing else refreshes the local node GPU,
+			// CPU and memory figures, so they stayed frozen at whatever they
+			// were when the daemon started. Observed on an RTX 3060: the card
+			// filled to 9.1 GB while the directory kept reporting the 0.12 GB
+			// it held at startup, with node-info serving the true value every
+			// two seconds the whole time. A peer converges because every
+			// browse event re-enriches it; self had no equivalent path.
+			d.publishSelf()
 		}
 	}
 }


Noah-Tervalon-Nvidia:
Thank you for your PR and interest in PAIR! In general we prefer pulls to pushes, so making the update in `applyInfo` would be the right direction.

canja006:
Happy to move it — but I can't find `applyInfo` in the tree, and I'd rather ask than guess at which function you mean.

I checked `main`, `origin/develop`, the full history (`git log --all -S`), every file type, case-insensitively: no match. The `apply*` functions that do exist are all in other services — `applyDiscovery` (cluster-manager, engine-manager, errors), `applyMembers`/`applyTombstones`/`applyRemovalProofs` (roster), `applyNodesChanged`/`applyUpsert`/`applyTelemetry` (job-scheduler) — none in `nvpair-node-scanner`.

The closest thing to an "apply info" step in the scanner is the tail of `enrichInfoCandidates`:

```go
node.GPUs = info.GPUs
node.CPU = info.CPU
node.Memory = info.Memory
```

reached through `enrich()` (all candidate IPs) or `enrichAt()` (one host). If that's what you meant, I'll work there.

One thing worth separating out in the push/pull framing, in case it changes the answer: `publishSelf` already pulls. `publishSelfLocked` builds the record, calls `d.enrichAt(&node, loopbackHost)` — the same node-info fetch a peer gets — and only then does `d.dir.upsert(node)` and `d.emit(...)`. So the tick in this PR does fetch over loopback; what it *also* does is republish the whole record, and I assume that's the part you'd rather not have on a timer.

If so, the narrower change is to re-run only the apply step for our own entry: fetch node-info over loopback on the refresh tick and update the existing directory node's `GPUs`/`CPU`/`Memory` in place, no `upsert`, no `emit`. That keeps it a pull and touches three fields instead of republishing. Say the word and I'll rewrite it that way.

For context on why self is the odd one out: `onBrowse` returns early when the event's UUID is our own, so self never gets the per-browse re-enrichment every peer gets, and `refreshPeersLoop` refreshes advertised addresses, the mesh, models and cluster identity — nothing that touches self's hardware figures. Hence the stale numbers in #4.

If `applyInfo` lives on a branch that isn't published, just say so and I'll match its shape once it lands.

PR #9, opened 2026-09-04
Add vLLM as a third inference engine

cguldogan:
## Description

PAIR supports two inference engines, Ollama and LM Studio. This adds **vLLM** as a third, on the same footing: detected or adopted, started and stopped, advertised over discovery, routed to, scheduled, and shown in the desktop and terminal interfaces.

User-visible outcome:

- vLLM appears in Engine settings on Linux with install, start, stop, restart, uninstall, and port configuration. On macOS and Windows it is listed but not installable, because vLLM publishes no GPU build there.
- An already running vLLM on the configured port (a `vllm serve …` or the `vllm/vllm-openai` container) is **adopted** rather than restarted. That is the normal path.
- A vLLM process serves one model chosen before it starts, so Engine settings gains a **Model to serve** field (a Hugging Face id). Starting vLLM with no model configured fails with a message pointing at the setting instead of a launch error. There are no pull, load, eject, or delete operations.
- vLLM shares the existing OpenAI-compatible proxy with LM Studio: one listener, one port, both engines' models in a single `/v1/models`. A peer's request is routed to whichever local engine owns the requested model.
- Manually added nodes are probed for vLLM on port 8000 (distinguished from other OpenAI-compatible servers by `GET /version`). The scheduler ranks vLLM nodes with the same load signal as the other engines.

## Scope

Included: engine manifest (`services/nvpair-engine-manager/manifests/vllm.json`, Linux amd64/arm64), the `runtime.model` / `runtime.extra_args` manifest fields and the persistent `engine:set-model` method, discovery key `vl`, the OpenAI-compatible proxy generalised to any OpenAI-compatible engine (binary name and relay namespace unchanged; README retitled), broker advertise loop and manual-node bridge, scheduler engine list, TUI labels, desktop engine type and capabilities, docs.

Excluded: a managed port facade for vLLM (it keeps port 8000 and is never moved), model download or deletion for vLLM, and any Windows or macOS install path.

**Breaking change for API consumers of `lmstudio-proxy`:** `node/add-manual` and `node/remove-manual` now require an `engine` field (`lmstudio` or `vllm`), hence its MAJOR bump.

Related: a sibling PR makes nodes added by address first-class peers on overlay networks. Both touch `services/nvpair-manual-nodes/manager.go` and `services/versions.json`; whichever merges second needs a small conflict resolution (vLLM becomes a third engine leg of the bare-host probe, and the higher version of each component is kept).

## Validation

Environment: macOS 15 arm64, Go 1.27, Node 23.

```
cd services && ./build.sh                       # 13 binaries, product 0.92.0
go test ./...   # nvpair-engine-manager, lmstudio-proxy, nvpair-ui-broker, shared, nvpair-manual-nodes, nvpair-job-scheduler, nvpair-tui
cd services/tests && go test ./...              # pass, 0 skips
cd desktop && npm run typecheck && npm run lint && npm run test:unit && npm run dead-code:check && npm run service-contracts:write && npm run service-contracts:check
node scripts/spdx-headers.mjs                   # 0 missing
```

All pass, except `nvpair-engine-manager TestUninstallTerminatesRunningInstance`, which fails identically on a pristine `main` on this host (it cannot resolve the image path of the process holding a port) and is unrelated.

End-to-end: a stub serving vLLM's `/health`, `/version`, `/v1/models`, and `/v1/chat/completions` was adopted by the real broker (`engine:get-installed` reported it running on its port with the configured model), the OpenAI proxy listed its model and completed a chat request, and the workload was tagged `vllm`. Adoption of a real two-node vLLM (`--tensor-parallel-size 2 --nnodes 2`) on Linux arm64 has also been exercised. The manifest's `install` block and the 30-minute readiness budget have not been run on real Linux GPU hardware.

## Risk

- Compatibility: the `engine` field on the proxy's manual-node methods is required (MAJOR bump on `lmstudio-proxy`). `EngineStatus` gains an optional `model`. Everything else is additive.
- Security: unchanged model. The managed manifest starts vLLM with `--host 127.0.0.1`; an adopted instance keeps whatever bind it was started with, exactly as an adopted Ollama does.
- Packaging: no new binary. Manifests are embedded.

Versions: `lmstudio-proxy` 0.16.2→1.0.0, `nvpair-engine-manager` 0.17.4→0.18.0, `nvpair-ui-broker` 0.40.2→0.41.0, `nvpair-manual-nodes` 0.11.1→0.12.0, `nvpair-job-scheduler` 0.4.1→0.5.0, patch bumps for `nvpair-node-scanner`, `nvpair-errors`, `nvpair-workload-manager`, `nvpair-tui`; product 0.91.7→0.92.0.

## Checklist

Related: #10 (touches the same manual-nodes file; see the note under Scope) and #11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ one, and both report live GPU and memory use throughout.
 | **Architectures** | x64 and arm64 on all three. Windows on ARM is experimental. |
 | **Installers** | Windows `.exe`; Linux `.deb`; macOS `.dmg`. On other Linux distributions, [build from source](docs/building.mdx). |
 | **Mixing nodes** | Windows, Linux, and macOS nodes can all be paired with each other |
-| **Inference engines** | Ollama and LM Studio |
+| **Inference engines** | Ollama, LM Studio, and vLLM (Linux) |
 
 **PAIR running on a machine does not mean an engine will.** PAIR itself runs on
 any supported Windows, Linux, or macOS machine. Each engine sets its own requirements
@@ -100,7 +100,8 @@ you want by its full filename instead.
   status there.
 
 - **Get an engine running.** On the node's card, open **Engine settings** and
-  select **Install** next to Ollama or LM Studio. PAIR downloads and sets the
+  select **Install** next to Ollama, LM Studio, or (on Linux) vLLM. PAIR
+  downloads and sets the
   engine up for you, so nothing needs to be in place beforehand. If PAIR already
   found an engine you installed yourself, start that one instead.
 
@@ -149,7 +150,8 @@ The reply is ordinary OpenAI-shaped JSON, abbreviated here:
 }
 ```
 
-If you changed a port, or you are using LM Studio rather than Ollama, copy the
+If you changed a port, or you are using LM Studio or vLLM rather than Ollama,
+copy the
 URL from **Endpoints → API endpoints** instead of assuming the one above.
 
 That is a single machine working. To route across machines, pair a second one
--- a/desktop/docs/services-api.md
+++ b/desktop/docs/services-api.md
@@ -123,6 +123,7 @@
 | `engine:remote-stop` | request (we call) | ✅ yes |
 | `engine:remote-unload-model` | request (we call) | ✅ yes |
 | `engine:restart` | request (we call) | ⚠️ not called |
+| `engine:set-model` | request (we call) | ✅ yes |
 | `engine:set-port` | request (we call) | ✅ yes |
 | `engine:start` | request (we call) | ✅ yes |
 | `engine:status` | request (we call) | ✅ yes |
--- a/desktop/docs/services-backend.md
+++ b/desktop/docs/services-backend.md
@@ -22,7 +22,7 @@ broker supervises every worker and relays its control plane.
 | ------------------------- | --------------------------------------------------------- |
 | `nvpair-ui-broker`        | Worker supervision and relay                              |
 | `ollama-proxy`            | Ollama-compatible routing proxy with cluster-mTLS ingress |
-| `lmstudio-proxy`          | LM Studio routing proxy with cluster-mTLS ingress         |
+| `lmstudio-proxy`          | OpenAI-compatible routing proxy (LM Studio, vLLM) with cluster-mTLS ingress |
 | `nvpair-node-scanner`     | Discovery and node announcement                           |
 | `nvpair-node-info`        | Node metadata and telemetry                               |
 | `nvpair-manual-nodes`     | User-managed node entries                                 |
@@ -90,8 +90,10 @@ engine, workload, cluster, and error relays. The bridge then emits renderer push
 events from backend notifications.
 
 Connector readiness follows the broker contract: `app:ready` establishes the
-service connection, while Ollama and LM Studio proxy readiness remains an
-asynchronous capability signal. Personal AI Router waits up to the canonical
+service connection, while the Ollama and OpenAI-compatible proxy readiness
+signals remain asynchronous capability signals. The OpenAI-compatible proxy
+fronts LM Studio and vLLM together, so one `lmstudio-proxy:ready` records the
+port for both engines. Personal AI Router waits up to the canonical
 startup deadline in `src/shared/constants/modular-runtime.ts` for
 `app:ready`; an outright failure or stalled broker startup is surfaced in
 Settings > Service with retry and log access. If a stalled broker reports ready
@@ -266,7 +268,8 @@ LAN-reachable. Each node fronts its engine with its `ollama-proxy` /
 `lmstudio-proxy`, whose LAN ingress is gated by cluster mTLS: only a pinned
 cluster member can send it work. Discovery advertises the promoted **proxy**
 port (never the engine port), and the broker hands the private loopback engine to
-the local proxy via `node/set-local-backend`. Every cluster-scoped worker derives
+the local proxy via `node/set-local-backend`, which is keyed per engine so the
+one OpenAI-compatible proxy can hold an LM Studio and a vLLM backend at once. Every cluster-scoped worker derives
 its own membership from the cluster directory continuously, so a proxy's mTLS
 ingress tracks a create, join, or leave with no worker restart. Loopback
 plaintext requests from local clients are unaffected. This is entirely
--- a/desktop/docs/services-parity.md
+++ b/desktop/docs/services-parity.md
@@ -24,6 +24,7 @@ history.
 | Manual nodes               | Complete with local persistence | Broker owns probing and proxy registration; Electron persists entries for replay                                                                |
 | Ollama routing             | Complete                        | Broker relay and backend scheduler drive proxy routing                                                                                          |
 | LM Studio routing          | Complete                        | Parallel broker relay and scheduler path                                                                                                        |
+| vLLM routing               | Complete on Linux               | Shares the LM Studio relay, the OpenAI-compatible proxy, and the same scheduler path                                                            |
 | Local engine lifecycle     | Complete                        | Install, start, stop, uninstall, update, and port configuration                                                                                 |
 | Remote engine lifecycle    | Partial                         | Remote install, start, stop, status, and model pull are supported                                                                               |
 | Engine models              | Partial                         | Core list, pull, load, unload, and supported delete actions are wired                                                                           |
@@ -102,7 +103,9 @@ they survive worker restarts.
 Both text-engine proxies are broker-owned and cluster-aware:
 
 - `ollama-proxy` serves the Ollama-compatible surface;
-- `lmstudio-proxy` serves the LM Studio/OpenAI-compatible surface.
+- `lmstudio-proxy` serves the OpenAI-compatible surface for every
+  OpenAI-compatible engine — LM Studio and vLLM share the one listener, and a
+  node record can advertise `lm` and `vl` at the same port.
 
 Routing precedence is manual selection, scheduler priority, then deterministic
 proxy ordering. Personal AI Router leaves proxies in automatic mode.
@@ -197,8 +200,16 @@ Personal AI Router uses:
 - Ollama `run_model`, `unload_model` (`keep_alive: 0`), and `delete_model`;
 - LM Studio `load_model`, `unload_model`, and `delete_model` (`remove_path`).
 
-Both engines expose Load, Eject, and Delete in the model manager when the
-backend action exists. Keep-alive / expiry controls remain unsupported.
+Ollama and LM Studio expose Load, Eject, and Delete in the model manager when
+the backend action exists. Keep-alive / expiry controls remain unsupported.
+
+vLLM has no model operations at all: it declares only `list_models`,
+`loaded_models`, and `chat`, because a vLLM process serves exactly one model
+chosen at launch and it neither downloads nor deletes weights on request. Its
+capability entry therefore sets `hasDeleteModel: false`, `hasEject: false`,
+`hasExpiry: false`, and `modelOpsWhenStopped: false`, and adds
+`hasServedModel: true` — the model to serve is a *setting* (`engine:set-model`),
+not a model action, and changing it restarts the engine.
 
 LM Studio's `delete_model` declares `restart_after`, so the engine manager
 restarts a running LM Studio once the files are removed — its `/v1/models` is
--- a/desktop/src/electron/service-bridge/empty-handlers.ts
+++ b/desktop/src/electron/service-bridge/empty-handlers.ts
@@ -507,6 +507,24 @@ function routeEngineManagerCommand(payload: WsInvokeRequest<'engine:command'>): 
         case 'toggle':
             void toggleLocalEngine(engine, payload.engineType)
             break
+        case 'setServedModel':
+            // vLLM serves one model per process, so the model is a start-time
+            // setting rather than a model operation. engine:set-model persists it
+            // as a manifest override — the same mechanism engine:set-port uses —
+            // and restarts a running engine onto it. Local only: the backend
+            // exposes no remote served-model control.
+            supervisor.sendProcess(
+                'broker',
+                'engine:set-model',
+                { engine, model: payload.model ?? '' },
+                failAction('set the served model', {
+                    nodeId: payload.nodeId,
+                    engineType: payload.engineType,
+                    modelName: payload.model
+                }),
+                true
+            )
+            break
         case 'setPorts':
             // Both ports persist on this node: the engine HTTP server port via
             // engine:set-port (manifest override), the proxy port via the broker's
--- a/desktop/src/electron/service-bridge/modular-state.ts
+++ b/desktop/src/electron/service-bridge/modular-state.ts
@@ -40,13 +40,23 @@ type BrokerNodeSource = ProxyNodeSource | 'broker'
  * Engines surfaced by the broker's proxy plane. Other engine-manager engines
  * are not currently routed across nodes.
  */
-export type ProxyEngine = Extract<EngineType, 'ollama' | 'lm-studio'>
-export const PROXY_ENGINES: readonly ProxyEngine[] = ['ollama', 'lm-studio']
+export type ProxyEngine = Extract<EngineType, 'ollama' | 'lm-studio' | 'vllm'>
+export const PROXY_ENGINES: readonly ProxyEngine[] = ['ollama', 'lm-studio', 'vllm']
 
-/** Map a proxy node source onto the engine it describes. */
-const PROXY_SOURCE_ENGINE: Record<ProxyNodeSource, ProxyEngine> = {
-    'ollama-proxy': 'ollama',
-    'lmstudio-proxy': 'lm-studio'
+/**
+ * The engines each proxy fronts. `lmstudio-proxy` is the OpenAI-compatible
+ * router and serves every engine that speaks that API, so a node it reports is
+ * not necessarily LM Studio — which engine it actually runs comes from the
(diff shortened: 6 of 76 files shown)


vedcsolution:
+1 — this would be a big deal for DGX Spark owners specifically.

Our setup: a small home/lab cluster of NVIDIA GB10 (DGX Spark) nodes paired in PAIR, each serving production models via `vllm serve` on :8000 (OpenAI-compatible, e.g. NVFP4-quantized Qwen and GLM MoE models with long context). Ollama and LM Studio don't really cover these: we need the vLLM-specific runtime options (quantization formats like NVFP4, speculative decoding, PLE/offload patches, multi-node TP) that aren't exposed through the other engines.

The mismatch today is that PAIR sees and manages the nodes (pairing, health, workload signals all work great) but the engine-manager only probes the fixed Ollama/LM Studio ports, so a healthy `vllm serve` is completely invisible — the Overview shows all nodes with no downloadable engines and no way to route to the models that actually live on them. We currently keep a separate LiteLLM gateway in front of the vLLM endpoints, which works but means PAIR can't do what it's uniquely good at: cluster-aware routing with node GPU/memory telemetry.

The issue body's "adopt an already-running vLLM on the configured port" framing is exactly what we'd use — we don't need PAIR to install or lifecycle-manage vLLM (systemd/docker handles that), just detect `/v1/models` on a configurable port per node, advertise the models over discovery, and route to them through the OpenAI-compatible endpoint. That "adopt, don't install" subset alone would unblock mixed clusters like ours.

Happy to test a preview build on a 6-node all-GB10 cluster if that's useful.

cguldogan:
> +1 — this would be a big deal for DGX Spark owners specifically.
> 
> Our setup: a small home/lab cluster of NVIDIA GB10 (DGX Spark) nodes paired in PAIR, each serving production models via `vllm serve` on :8000 (OpenAI-compatible, e.g. NVFP4-quantized Qwen and GLM MoE models with long context). Ollama and LM Studio don't really cover these: we need the vLLM-specific runtime options (quantization formats like NVFP4, speculative decoding, PLE/offload patches, multi-node TP) that aren't exposed through the other engines.
> 
> The mismatch today is that PAIR sees and manages the nodes (pairing, health, workload signals all work great) but the engine-manager only probes the fixed Ollama/LM Studio ports, so a healthy `vllm serve` is completely invisible — the Overview shows all nodes with no downloadable engines and no way to route to the models that actually live on them. We currently keep a separate LiteLLM gateway in front of the vLLM endpoints, which works but means PAIR can't do what it's uniquely good at: cluster-aware routing with node GPU/memory telemetry.
> 
> The issue body's "adopt an already-running vLLM on the configured port" framing is exactly what we'd use — we don't need PAIR to install or lifecycle-manage vLLM (systemd/docker handles that), just detect `/v1/models` on a configurable port per node, advertise the models over discovery, and route to them through the OpenAI-compatible endpoint. That "adopt, don't install" subset alone would unblock mixed clusters like ours.
> 
> Happy to test a preview build on a 6-node all-GB10 cluster if that's useful.

full implementation is here I would love to see it :)

https://github.com/cguldogan/Personal-AI-Router/releases

arcaven:
I built and ran the suite at a4b96a3 on macOS arm64, since this adds an engine and a second platform data point seemed worth having. The vLLM work passes; the two failures I hit are pre-existing on main and are not from this PR.

Environment: darwin/arm64 (Apple Silicon), Go 1.26.5, Node 25.9.0, clean tree at a4b96a3.

**Desktop** (`make test-desktop`): 39 files, 220 tests, all pass, including the new `vllm-engine.test.ts` (6) and `openai-proxy-engines.test.ts` (6).

**Go**: `make test-services` halts at the first failing module, so I ran each module separately. 18 modules, 16 pass, 2 fail:

| module | test | also fails at origin/main 13b6811 |
|---|---|---|
| `nvpair-engine-manager` | `TestUninstallTerminatesRunningInstance` | yes |
| `ollama-proxy` | `TestAliasSelfTargetMatchesBoundLoopbackAddressNotPortAlone` | yes |

Both reproduce at main with identical messages, so neither is caused by this change. `lmstudio-proxy`, which carries most of the new engine code plus `multiengine_test.go`, passes.

On those two, in case it saves anyone time:

- The alias test binds `127.0.0.2`. macOS configures only `127.0.0.1` on `lo0`, where Linux treats all of `127.0.0.0/8` as local, so the bind fails before the test's own logic runs. An `ifconfig lo0 alias` in the harness or a platform skip would cover it.
- `TestUninstallTerminatesRunningInstance` may reach a real macOS gap rather than a test-only one. `procImage` in `proc_unix.go` reads `/proc/<pid>/exe`, which macOS does not provide, so it returns `""`; `isOurEngineImage` returns false for an empty image; and the orphan-reclaim branch in `doStop` is skipped as a result. That would leave the path described in its own comment, letting a user's OFF take effect "instead of being refused forever", unreachable on macOS. That is my reading of the source rather than something I instrumented. Happy to file it separately if useful.

Two small notes on the manifest, neither of them asks:

- `manifests/vllm.json` declares `linux/amd64` and `linux/arm64` only, which I read as deliberate. It does mean an Apple Silicon tester can exercise the Go and TS layers but not the runtime path.
- The manifest exposes `/v1/models` and `/v1/chat/completions`. vLLM also serves the Anthropic Messages API (`vllm/entrypoints/anthropic/api_router.py`), which is the subject of #16. Nothing this PR needs to take on; noting it only because the two touch.

arcaven:
Filed the macOS piece separately as #17, since it reproduces at origin/main and is a different root cause from anything here. Nothing in it blocks this PR.

arcaven:
Small one from the same pass: three files in this PR are not gofmt-clean.

```
$ git checkout a4b96a3 && gofmt -l services/ scripts/
services/nvpair-engine-manager/registry.go
services/nvpair-node-info/stats_windows.go
services/shared/noderec/noderec.go
services/shared/splitlisten/splitlisten_test.go
services/tests/model_routing_interop_test.go

$ git checkout 13b6811 && gofmt -l services/ scripts/     # merge base
services/nvpair-node-info/stats_windows.go
services/shared/splitlisten/splitlisten_test.go
```

So `registry.go`, `noderec.go`, and `model_routing_interop_test.go` are the three this branch introduces; the other two are already on main and are not yours. In `registry.go` it is struct tag alignment in `Runtime`, where the new `ExtraArgs` field widens the column and the fields after it were not realigned. `gofmt -w` on the three fixes it.

Two other notes from testing this branch on Apple Silicon, so they are not mistaken for anything here:

- The `lmstudio-proxy` failure I mentioned earlier turned out to be machine state, not this PR and not main. It is #48.
- I opened #47 for the `nvpair-engine-manager` test that fails on macOS at main, which is the one from my first comment.

jlacroix82:
I've built on this PR at `a4b96a3` to add SGLang alongside vLLM for our existing cluster deployments. The work is in my fork, keeping this PR's vLLM implementation and shared OpenAI proxy intact:

https://github.com/jlacroix82/Personal-AI-Router/tree/feat/vllm-sglang

Extension commit: https://github.com/jlacroix82/Personal-AI-Router/commit/58508c665f4bcd9e58da814a8f0a7d443ab1c26c

The extension adds:

- SGLang desktop engine registration, node visibility, model inventory, and proxy attribution.
- An operator-installed SGLang manifest, default port 30000, and adoption of an already-running process. No new installer or runtime dependency is introduced.
- The `sg` discovery key through the existing OpenAI proxy, separate local engine state, and scheduler attribution.
- Bare-host SGLang detection using `/get_model_info` plus `/v1/models`, with `sglang_*` status fields relayed by the broker.
- Local port/model configuration before detection, so an existing server on a non-default port can be connected from Engine settings.

I also fixed the broker's `manualToEnriched` combined model list to include vLLM and SGLang; previously it combined only Ollama and LM Studio, although the per-engine map included vLLM.

Validation on Linux ARM64, Node 26.5.0, Go 1.26.5:

- Desktop typecheck: pass; unit tests: 40 files / 222 tests pass.
- Lint: no errors, one pre-existing formatting warning in `node-info-poller.ts`; dead-code and service-contract checks pass.
- `go test ./...` passes in shared, engine-manager, OpenAI proxy, UI broker, manual-nodes, and job-scheduler. Added tests cover SGLang adoption, identity/inventory probing, proxy advertisement and withdrawal, and HTTP ingress to separate vLLM/SGLang servers.
- All 13 service binaries and the Linux ARM64 Electron preview build/package successfully.
- Extended the existing cross-process strict-routing test with vLLM and SGLang cases; execution skipped because the host's PAIR default ports were occupied. This is not counted as a pass for that scenario.

This is stub-backed validation, not a claim of a completed production-cluster rollout. Native single-model controls are retained; externally managed processes remain under their supervisor. Authenticated engine endpoints and fully dynamic arbitrary-engine registration are still outside this change.

The adapter repository now also has a separate SGLang manifest/configuration, non-conflicting wrapper/child ports, two passing mock-adapter tests, and GUI setup instructions:
https://github.com/jlacroix82/pair-multi-engine

Happy for the SGLang follow-up to stay separate from #9. The branch is available to review or cherry-pick once the vLLM foundation is settled.

NV-sschneider:
Thanks all for your contributions to the project!  We are just getting back from the long weekend and will be reviewing this PR and all the others as well. 

We are very excited about vLLM support!

Hang tight. :)

cguldogan:
> I've built on this PR at `a4b96a3` to add SGLang alongside vLLM for our existing cluster deployments. The work is in my fork, keeping this PR's vLLM implementation and shared OpenAI proxy intact:
> 
> https://github.com/jlacroix82/Personal-AI-Router/tree/feat/vllm-sglang
> 
> Extension commit: [jlacroix82@58508c6](https://github.com/jlacroix82/Personal-AI-Router/commit/58508c665f4bcd9e58da814a8f0a7d443ab1c26c)
> 
> The extension adds:
> 
> * SGLang desktop engine registration, node visibility, model inventory, and proxy attribution.
> * An operator-installed SGLang manifest, default port 30000, and adoption of an already-running process. No new installer or runtime dependency is introduced.
> * The `sg` discovery key through the existing OpenAI proxy, separate local engine state, and scheduler attribution.
> * Bare-host SGLang detection using `/get_model_info` plus `/v1/models`, with `sglang_*` status fields relayed by the broker.
> * Local port/model configuration before detection, so an existing server on a non-default port can be connected from Engine settings.
> 
> I also fixed the broker's `manualToEnriched` combined model list to include vLLM and SGLang; previously it combined only Ollama and LM Studio, although the per-engine map included vLLM.
> 
> Validation on Linux ARM64, Node 26.5.0, Go 1.26.5:
> 
> * Desktop typecheck: pass; unit tests: 40 files / 222 tests pass.
> * Lint: no errors, one pre-existing formatting warning in `node-info-poller.ts`; dead-code and service-contract checks pass.
> * `go test ./...` passes in shared, engine-manager, OpenAI proxy, UI broker, manual-nodes, and job-scheduler. Added tests cover SGLang adoption, identity/inventory probing, proxy advertisement and withdrawal, and HTTP ingress to separate vLLM/SGLang servers.
> * All 13 service binaries and the Linux ARM64 Electron preview build/package successfully.
> * Extended the existing cross-process strict-routing test with vLLM and SGLang cases; execution skipped because the host's PAIR default ports were occupied. This is not counted as a pass for that scenario.
> 
> This is stub-backed validation, not a claim of a completed production-cluster rollout. Native single-model controls are retained; externally managed processes remain under their supervisor. Authenticated engine endpoints and fully dynamic arbitrary-engine registration are still outside this change.
> 
> The adapter repository now also has a separate SGLang manifest/configuration, non-conflicting wrapper/child ports, two passing mock-adapter tests, and GUI setup instructions: https://github.com/jlacroix82/pair-multi-engine
> 
> Happy for the SGLang follow-up to stay separate from #9. The branch is available to review or cherry-pick once the vLLM foundation is settled.

I saw this after the SGLang PR had already been submitted. You could open a PR as well maybe your version has a better chance of being accepted?

jlacroix82:
Follow-up on the vLLM/SGLang GUI branch: testing against an existing native deployment turned up a few things that needed tightening up.

Changes are in [21a363d](https://github.com/jlacroix82/Personal-AI-Router/commit/21a363d) on [feat/vllm-sglang](https://github.com/jlacroix82/Personal-AI-Router/tree/feat/vllm-sglang):

- External processes are labeled as externally managed. Desktop and TUI no longer offer lifecycle controls for them; declared deployments also reject lifecycle/model/port mutations in the Go service.
- Native adoption checks engine identity as well as health (`/version` for vLLM, `/get_model_info` for SGLang). The broker no longer guesses a native endpoint when engine-manager is unavailable.
- An operator-maintained `engine-deployments.json` describes API heads, workers and capabilities. Workers aren't separate routes for that engine, and their GPU pressure contributes to the head's scheduling load. This is explicit configuration, not automatic topology discovery or PAIR-managed tensor parallelism.
- Embedding-only deployments retain their model inventory but reject chat/completion routing. Capability metadata reaches inference-dispatcher and the generation demo.
- The GUI model parser now accepts OpenAI `data[].id` inventories. It previously understood Ollama and native LM Studio responses but missed the native vLLM/SGLang shape.
- Engine settings have a connection/status refresh control, and model catalogs distinguish available models from confirmed loaded models.
- Fixed a shared-proxy scheduling issue: vLLM, SGLang and LM Studio must not race to overwrite the same proxy priority store.

Validation: 224 desktop unit tests pass; typecheck, contracts and dead-code checks pass. Lint has no errors (one existing formatting warning). Go suites pass for shared, engine-manager, broker, OpenAI proxy, scheduler and TUI; inference-dispatcher tests and both adapter tests pass. ARM64 packaging succeeds. A live native vLLM embedding endpoint is detected as external/loaded, retains embedding metadata through PAIR, and rejects a chat request at the proxy. The model container was not restarted.

SGLang adoption/routing has simulated-server coverage, not a new live SGLang GPU benchmark. Remote fleet rollout and controlled throughput benchmarking are still separate steps. Participating nodes need matching preview services and deployment policy; stock PAIR does not acquire this behavior from a manifest alone.

[Setup and limits](https://github.com/jlacroix82/Personal-AI-Router/blob/feat/vllm-sglang/docs/external-deployments.mdx). The optional adapter repo also has updated identity overrides and setup notes in [a40049a](https://github.com/jlacroix82/pair-multi-engine/commit/a40049a).

Service versions: engine-manager 0.20.0, OpenAI proxy 1.2.0, scheduler 0.7.0, broker 0.42.1 and TUI 0.8.1. These changes are on my fork branch; this comment links the follow-up rather than modifying the PR author's source branch.

jlacroix82:
Another follow-up from testing this against the running Spark cluster: the UI was mixing up “PAIR owns this process” with “this process is running.” It also treated distributed workers with no HTTP server as stopped.

The fix is on my fork: [ff26f5b](https://github.com/jlacroix82/Personal-AI-Router/commit/ff26f5b), branch `feat/vllm-sglang`.

- External engines keep a checked, read-only switch when running, instead of losing the switch entirely.
- Node rows identify the deployment and head/worker role. Settings separate runtime activity from API readiness.
- An optional `containers` map in `engine-deployments.json` gives each member an exact local Docker container to inspect. The check is bounded and read-only; it doesn't depend on which vLLM/SGLang image or launcher CLI is used.
- Workers can be running with no local API. They remain excluded from inference routing. Missing/unobservable worker state and failed peer status polls show Unknown, not Off.

I rolled the matching PAIR services and policy onto eight cluster devices hosting three distributed vLLM deployments, plus the local embedding host. The live GUI reports all nine vLLM switches checked and read-only, with three cluster API heads and five workers. Model-container IDs, start times and PIDs stayed unchanged. No model containers were restarted and no inference benchmark load was sent.

Validation: 226 desktop unit tests; shared, engine-manager, broker, scheduler, OpenAI proxy and TUI Go suites; typecheck, contracts, dead-code and SPDX checks. Lint has no errors and one existing unrelated formatting warning. Two ingress tests now isolate their config so a developer's actual deployment policy can't change their expected results.

One limit worth being explicit about: Docker “running” describes the container, not the health of every child process. Head API readiness is still checked separately. Non-Docker workers remain unknown until a suitable observer exists. SGLang follows the same code path and has test coverage, but I don't currently have a live SGLang deployment for a GPU integration run.

Setup and wire-field details: [external deployments](https://github.com/jlacroix82/Personal-AI-Router/blob/feat/vllm-sglang/docs/external-deployments.mdx). This is a follow-up commit for the PR author to pick up; it doesn't change the PR's source branch directly.

s33thr33p10:
## Coordination request: security follow-up to this PR

Thank you for the vLLM work in #9. We built a focused follow-up that keeps vLLM adopt-and-route-only: PAIR can discover and route to it, but cannot install, start, stop, restart, reconfigure, move ports, mutate models, or uninstall the external server.

To preserve attribution and avoid opening a competing vLLM PR, our preferred options are:

1. send the focused security changes to `cguldogan:feat/vllm-engine`; or
2. wait for #9 to merge, then open an incremental security follow-up.

The follow-up is based directly on `a4b96a3`, and the first six commits remain exactly those from #9. We will publish only the source changes required for the upstream project; private validation methods, infrastructure, evidence, and environment details are out of scope.

We also need maintainer guidance on the repository's SPDX policy before publishing it. Two genuinely new shared authority files currently carry contributor notices rather than NVIDIA's canonical notice. We do not want to claim NVIDIA copyright without an authorized ownership basis merely to satisfy the checker. Should external Apache-2.0 contributor notices receive a documented exception/review, or does NVIDIA's contribution process provide the basis for the canonical notice?

Please let us know which contribution topology and notice treatment you prefer. We will keep the follow-up unpublished until that is clear.

Noah-Tervalon-Nvidia:

PR #10, opened 2026-09-04
Make a node added by address a first-class peer (overlay networks such as Tailscale)

cguldogan:
## Description

PAIR finds peers by multicast DNS, which never leaves a local link. Two machines joined only by an overlay network (a Tailscale tailnet, a WireGuard tunnel, a routed link) were never going to discover each other, and adding such a peer by address did not work either: PAIR probed the peer's Ollama and LM Studio ports in plaintext, and on a machine running PAIR those ports are proxy front doors that refuse plaintext from anything but their own loopback. Every probe was refused, so a healthy peer reported no engines and nothing could be routed to it.

This makes a PAIR node added by address behave exactly like one found on a LAN: same card, engines, models, and telemetry, with inference routed to it over the same certificate-pinned mutual TLS.

- `nvpair-node-info` reports this node's service map on `/v1/node-info`, fed by the broker.
- `nvpair-manual-nodes` asks node-info first; a node that answers with a service map is a PAIR node and is never probed on its engine ports. Its inventory is read from its engine manager over cluster mTLS. A node without one is a bare inference host and keeps the original plaintext engine probes.
- The broker folds a manual PAIR node into the discovery directory, so proxies, scheduler, remote engine operations, and workload relay treat it as a discovered pinned peer.
- **Add node** accepts a host name (re-resolved on every probe), rejects `host:port` with an actionable message, and gains a **Service ports** section for a node whose services are not on the defaults
- `netpick` admits DNS names as dialable addresses. CGNAT (100.64/10) and overlay-interface demotion remain rankings, never exclusions, now pinned by tests.

## Scope

Included: the above, a cross-process test for a peer addehe desktop Add-node flow (create the node, then pair),`docs/remote-networks.mdx` with an ACL port list and troubleshooting, and README and SECURITY wording about the network boundary.

Excluded: automatic discovery of tailnet peers (planned follow-up), and any code that knows about a specific VPN product. Nothing here is
Tailscale-specific; it makes PAIR work on any routed netw

Related: a sibling PR adds vLLM as a third engine. Both tal-nodes/manager.go` and `services/versions.json`;whichever merges second needs a small conflict resolution (vLLM becomes a third engine leg of the bare-host probe, and the higher version of
each component is kept).

## Validation

Environment: macOS 15 arm64, Go 1.27, Node 23; two Linux ver a Tailscale tailnet.

```
cd services && ./build.sh
go test ./...   # shared, ollama-proxy, lmstudio-proxy, nnual-nodes, nvpair-ui-broker, nvpair-cluster-manager,nvpair-node-scanner, nvpair-errors, nvpair-workload-manager
cd services/tests && go test ./...              # 87 pass
cd desktop && npm run typecheck && npm run lint && npm run test:unit && npm run dead-code:check && npm run service-contracts:check
node scripts/spdx-headers.mjs
```

Pre-existing, environment-specific failures also present on `main`: `ollama-proxy TestAliasSelfTargetMatchesBoundLoopbackAddressNotPortAlone`
(needs a `127.0.0.2` loopback alias), and the two tests t-port file from the real user data directory after anothertest writes it.

Live check: two headless Linux nodes on a tailnet, each running PAIR from this branch, were added by address from a desktop node, paired with
the PIN exchange, and served inference from the desktop'sfore this change the same nodes appeared with no enginesand could not be paired.

## Risk

- Security: the trust model is unchanged. A manual PAIR node gets its models and routing only after pairing, exactly like a discovered one.
Plaintext engine probes still run for bare hosts, as befo
- Compatibility: `node/add` gains optional `ports`; `/v1/node-info` gains `services`; a `host:port` address, previously accepted and
permanently unreachable, is now rejected. Additive otherw
- Behavioural note: for a PAIR node, the broker issues a no-op `node/remove-manual` to the proxies on most probe cycles. It is cheap and
idempotent, but it runs inline on the manual-nodes reader
- Known limitation: if a peer's node-info stops while its proxies keep running, the node reverts to bare-host probing after about 30 s and the
peer logs a rejected plaintext request every 10 s until n

Versions: MINOR for `nvpair-node-info`, `nvpair-manual-no`ollama-proxy`, `lmstudio-proxy`; PATCH for`nvpair-node-scanner`, `nvpair-errors`, `nvpair-workload-manager`, `nvpair-cluster-manager`; product 0.91.7→0.92.0.

## Checklist

- [x] Every commit is signed off (`git commit -s`), certificate of Origin](https://developercertificate.org/).
- [x] I bumped any affected component in `services/versioer-visible changes above so they reach the release notes.

Related: #9 (touches the same manual-nodes file; see the note under Scope) and #11.


DustinTrap:
Tested the mixed-version case, since that is what anyone adopting this will live in for a while: one node on this branch, its peers on 0.1.1.

**Environment.** All thirteen services built from e6f8b80 (linux/amd64, Go 1.26, `CGO_ENABLED=0`), running as a `hostNetwork` pod on single-node OpenShift 4.22 with a fresh identity. Peers: a 0.1.1 headless Linux node with one GPU (LM Studio engine) and a 0.1.1 macOS desktop. The branch merges cleanly with #30 (both touch `netpick_test.go`, no conflict).

**node-info.** The branch node serves `/v1/node-info` on plaintext 14318 with the six-entry `services` map (`cl`, `ec`, `em`, `er`, `ni`, `wl`) before it is paired with anyone. The 0.1.1 peer's `/v1/node-info` has no `services` key.

**Branch node adds the 0.1.1 peer by address.** `nvpair-manual-nodes` logs `ollama_up=false node_info_up=true models=0 gpus=1` on every cycle, and the peer's `ollama-proxy` and `lmstudio-proxy` log `rejected non-loopback plaintext request; cluster peers must use mTLS` every 10 s. That is the bare-host fallback the comment in `manager.go` describes for "a PAIR node too old to report the set", working as written. `gpus=1` comes from the peer's node-info telemetry despite the bare-host classification.

**0.1.1 peer adds the branch node by address.** Identical to before this PR: `ollama_up=false node_info_up=true models=0`, since 0.1.1 never asks for the service map.

**Take-away.** The PR delivers once both ends run it; with a mixed fleet, the only visible change is the every-10-s rejection line on the older peer's proxies, which the description already notes for the node-info-down case and which applies equally to a not-yet-upgraded peer. A line in `docs/remote-networks.mdx` saying both nodes need this version or newer would save someone the same test.

Could not exercise the both-sides path here, since the other nodes run release builds. Happy to re-run when a release carries this; the pod setup for it is in #31.

ckelseynv:
Hi @cguldogan, thanks so much for your contributions! 

We are currently working on a Tailscale implementation from our roadmap and I'm looking into your PR right now. There are many things I'd love to bring in as is but may need to pull some out for a follow up PR. There is a bunch of work on our end regarding auto discovery, install, auth, etc that I plan on putting on top of your branch if that works for you?

cguldogan:
> Hi @cguldogan, thanks so much for your contributions!
> 
> We are currently working on a Tailscale implementation from our roadmap and I'm looking into your PR right now. There are many things I'd love to bring in as is but may need to pull some out for a follow up PR. There is a bunch of work on our end regarding auto discovery, install, auth, etc that I plan on putting on top of your branch if that works for you?

Of course. However you’d like to proceed, and whatever is easiest for you to manage, works for me.

Noah-Tervalon-Nvidia:

PR #14, opened 2026-09-04
fix: resolve 14 review issues across the service control plane

woodsonl:
## Summary

PAIR routes independent inference requests across local nodes. This PR resolves 14 issues found in a whole-repo review plus later adversarial and regression rounds across the service control plane, and adds coverage at each boundary the fixes touch.

**The 14 issues, by commit:**

1. **Stdio JSON-RPC read loops did not distinguish recoverable from terminal read errors** (broker + all 12 service modules): one malformed frame could spin or terminate the control-plane connection. DecodeError frames now skip-and-continue; EOF stays clean; terminal scanner/transport errors stop instead of spinning (`fcd03b3`).
2. **One slow worker relay head-of-line blocked the whole control plane** — broker relays ran inline on the read loop. Now a bounded 4-goroutine dispatch pool (`fcd03b3`).
3. **A stalled worker subscriber stalled every other subscriber** — relay delivery was synchronous. Now per-subscriber pump goroutines with a coalescing non-blocking `Deliver` (`fcd03b3`).
4. **Pairing PIN comparison now uses PBKDF2-HMAC-SHA256 stretching** (50k iterations, per-invite salt inside the EAP-MAC-covered ServerInfo) (`fcd03b3`).
5. **Pairing completion is now capped at 5 attempts** with invite teardown (`fcd03b3`).
6. **Terminal pairing signals (cancel/decline/fail/expire) are now authenticated with an HMAC** over the session's ephemeral Key Exchange secret (`fcd03b3`).
7. **Errors ingest now requires the envelope nodeId to match the mTLS-authenticated caller UUID**, with a 1 MiB body cap (`fcd03b3`).
8. **Proxies now enforce a deny-by-default origin allowlist** (`NVPAIR_PROXY_ALLOWED_ORIGINS`) and no longer echo arbitrary headers (`fcd03b3`).
9. **Inference requests now have a 32 MiB body cap**, returning 413 over the limit (`fcd03b3`).
10. **Engine stop now escalates** from graceful stop to SIGKILL/pgid after the manifest grace period (`fcd03b3`).
11. **Manifest installs now fail closed on unpinned downloads** unless `NVPAIR_ALLOW_UNPINNED_DOWNLOADS=1` (`fcd03b3`).
12. **Workload checkpoint re-marks the store dirty on write failure** so a failed snapshot is not treated as clean (`fcd03b3`).
13. **`safeHandle` no longer authorizes arbitrary renderer origins when `ELECTRON_RENDERER_URL` is empty** (`fcd03b3`).
14. **Cluster-manager session handling no longer double-unlocks `sess.mu`** on the over-completion-attempts branch and no longer leaves it locked on the cancel path, so unauthenticated POSTs can no longer panic or deadlock the manager (`fb2f81e`).

Follow-ups close the remaining gaps the rounds surfaced: lookalike-host/userinfo bypass of the dev-URL sender check (`9600675`), the same decode-error confusion in the broker's consumer side (`9600675`), and the relay `Deliver`-vs-`Unsubscribe` race (`9600675`).

## Intentionally in scope

Everything above, plus the regression tests listed under Validation, and the `services/versions.json` bumps required by the compiled-output changes (cluster-manager 1.1.6, engine-manager 0.17.6, ui-broker 0.40.4).

Intentionally out of scope: the `NVPAIR_SERVICE_*_PROXY_PORT` env overrides still skip the engine-port collision exclusion that the fallback path applies (`proxyport.go`/`lmstudioport.go`). They are operator/test-only knobs; tightening them touches the cross-process test harness and deserves its own change.

## Validation environment

- Linux, Go 1.25.0, Node.js per `desktop/package.json`
- Component suites with `-race`: `nvpair-cluster-manager`, `nvpair-ui-broker`, `eap-noob`, `ollama-proxy`, `lmstudio-proxy` — all pass
- Cross-process suite: `services/tests` (`go test ./...`) — passes
- Desktop: `npm run typecheck`, `npm run lint`, `npm run test:unit` (38 files), `npm run dead-code:check` — all pass
- `node scripts/spdx-headers.mjs` — 877 checked, 0 missing

New regression coverage at each fixed boundary:

- `desktop/tests/modular/safe-handle-sender.test.ts` — sender authorization: dev-URL exact and path-suffix match, `file://`, unset and empty env, lookalike host, userinfo spoofing, missing window (red/green against the fix)
- `services/nvpair-cluster-manager/pairing_signal_gate_test.go` — 401 gate for cancel/decline/expire with wrong-phase and wrong-invite tags MACed under the live session key, state survival; 429 rate limit including failed/incorrect-pin attempts and teardown; cancel without a signal key
- `services/nvpair-ui-broker/terminal_read_test.go` — read-loop contract: transport error terminal, EOF clean, decode error recoverable
- `services/nvpair-ui-broker/relay/relay_test.go` — trigger coalescing delivers latest state once; post-unsubscribe `Deliver` is silent
- `services/nvpair-ui-broker/ollamahost_test.go` — `resolveServicePorts` defaults, valid overrides, invalid values ignored
- `services/ollama-proxy/` and `services/lmstudio-proxy/` `ingress_test.go` — loopback cross-origin gate: unlisted origin 403 `origin-not-allowed`, allowlisted origin passes the gate
- `services/ollama-proxy/` and `services/lmstudio-proxy/` `body_limit_test.go` — body cap (over/at limit), model parse, end-to-end 413
- `services/eap-noob/ephemeral_key_test.go` — ephemeral key lifecycle: pre-exchange errors, server/peer agreement, copy semantics

## Compatibility notes

- No JSON-RPC method or payload shapes change; the fixes tighten enforcement of existing contracts. Paired-cluster and pairing flows are unaffected for legitimate participants.
- Legitimate pairing completion requests now always consume an attempt; a client that retried with empty bodies to probe state will now hit the limit — intended.
- The `safeHandle` dev-URL change only affects development-mode renderer IPC (`ELECTRON_RENDERER_URL`); packaged builds authorize via `file://` and are unchanged.
- The digest pins make installs fail closed if a registry serves artifacts that do not match the reviewed manifest — intended; updating an engine version requires updating the pin, which is the review point.

## Documentation changes

Service READMEs updated in the final commit to match the new behavior, where they previously described the old one:

- `services/ollama-proxy/README.md` (the CORS reference `lmstudio-proxy` links to): permissive wildcard grant → deny-by-default `NVPAIR_PROXY_ALLOWED_ORIGINS` allowlist, static preflight grant, `403 origin-not-allowed` gate
- `services/nvpair-engine-manager/README.md`: install "verify-if-pinned with a loud warning" → fail closed unless `NVPAIR_ALLOW_UNPINNED_DOWNLOADS=1`; stop "no timeout, no SIGKILL escalation" → manifest grace then forced kill
- `services/nvpair-cluster-manager/README.md` note: records the PIN stretching, authenticated terminal signals, and completion-attempt cap it describes

No user-facing workflow, configuration surface, endpoint, or default changes beyond the operator env vars above.

## Related issue

None — found and fixed through whole-repo review.


kjlubick:
> Read [SECURITY.md](https://github.com/NVIDIA/Personal-AI-Router/blob/main/SECURITY.md) before deploying it on an untrusted or shared network. Do not report vulnerabilities in a public issue.

Please note the above from the README. In addition, smaller PRs are easier to evaluate than large ones.

Noah-Tervalon-Nvidia:

kjlubick:
Please do not file security fixes like this (in the public). Follow the steps on https://github.com/NVIDIA/Personal-AI-Router/blob/develop/SECURITY.md

I did review the changes here and none seem actionable given our threat model and the changes that have recently landed on the `develop` branch

PR #18, opened 2026-09-05
Add adopt-only llama.cpp engine and distributed inference proxy

agenticocho:
## Summary

Add first-class, adopt-only llama.cpp support to Personal AI Router, using an existing local `llama-server` rather than installing or owning the runtime.

## Changes

- Add an adopt-only llama.cpp engine-manager manifest.
- Add `llamacpp-proxy`, an OpenAI-compatible distributed inference router with discovery, model-aware selection, failover, reservations, workload events, scheduler priorities, and cluster-mTLS ingress.
- Integrate the proxy into the UI broker lifecycle, JSON-RPC relay, health advertising, discovery registration (`ll`), log fan-out, and orderly shutdown.
- Add Desktop modular-binary, build-script, service-contract, and macOS firewall-helper parity.
- Bump `nvpair-ui-broker` from 0.40.2 to 0.41.0 and update generated and hand-maintained documentation.
- Prevent Electron's structured logger from raising uncaught `EPIPE` errors when a piped development launch loses its stdout consumer during shutdown.

## Validation

- Built all 14 service binaries successfully on Linux.
- `go test ./... -count=1` passed for `nvpair-ui-broker`.
- llama.cpp proxy package tests cover routing, failover, ingress, trust, persistence, priorities, reservations, transport pooling, and shutdown behavior.
- Desktop TypeScript typechecks passed.
- Desktop unit suite passed: 37 files, 208 tests.
- Build-script verification and service-contract checks passed.
- Runtime-tested with an adopted `llama-server` on `127.0.0.1:8080` and the broker-supervised proxy on `:8081`.
- Verified model discovery, successful chat completion, llama.cpp workload attribution, backend outage/recovery, proxy teardown, and no orphan worker processes.
- Verified localhost-only backend CORS in the runtime test configuration.

## Notes

The engine remains adopt-only: this change does not install llama.cpp or take ownership of an externally started server. API-key-protected adopted servers still require the calling client to provide the matching authorization header; backend credential injection is intentionally not added here.


Noah-Tervalon-Nvidia:
Thank you for opening this up! We're working on bringing in llama.cpp as well and want to make sure we bring it in fully and correctly.

agenticocho:
> Thank you for opening this up! We're working on bringing in llama.cpp as well and want to make sure we bring it in fully and correctly.

Thanks, Noah — great to hear llama.cpp is already on your roadmap.

We’ve continued validating the integration beyond the current PR head and have
useful findings around adopt-only lifecycle handling, authenticated local
backends, exact model-capability propagation, OpenAI-compatible /v1/models
parsing, failover safeguards, and macOS runtime behavior.

The current PR branch remains unchanged while we keep that follow-up work
isolated and reviewed. If useful, we’d be happy to share the focused test cases
and implementation findings, or compare notes against NVIDIA’s planned design
before proposing any update to the PR.

Issue #7, opened 2026-09-04
[Feature]: External / heterogeneous scheduling hooks and richer placement signals


guydax:
### Area

Routing and scheduling

### User problem

PAIR provides a strong transport, discovery, and inference data plane for distributed local inference, but heterogeneous clusters need richer placement signals than model ownership, queue depth, and coarse GPU utilization alone.

Today, an external scheduler cannot safely influence placement for one request without changing a persistent or proxy-wide pin, which can introduce races under concurrent traffic. Nodes that own the same model can also have very different practical runtime envelopes: characterized context limits, memory requirements, full-GPU versus CPU/offload behavior, model-load time, prompt/generation throughput, and reliability.

Requests themselves vary substantially in cost based on prompt length, requested context, expected output, model size, and modality. PAIR also has useful model-state visibility, but warm-model affinity and host-wide pressure from other compute workloads are not fully represented in routing decisions. Finally, operators need explainable placement metadata without exposing prompt content.

### Desired outcome

PAIR should optionally support the following while preserving its native distributed scheduler as the self-contained default and fallback:

1. **Request-scoped placement hints** such as a preferred node or ranked candidate list. PAIR would continue to enforce model ownership, availability, trust, and normal safety checks.
2. **Per-model/per-node runtime profiles** learned empirically, including characterized context sizes, VRAM/RAM requirements, GPU/offload behavior, load time, prompt TPS, generation TPS, and stability history.
3. **Weighted request-cost estimation** using inputs such as prompt tokens, requested context, expected output length, model size, modality, and historical model/node cost.
4. **Warm-model affinity** when a suitable model is already resident and resource pressure does not favor another node.
5. **An optional external placement-authority hook**, with automatic fallback to PAIR's native scheduler if that authority is unavailable.
6. **Host-wide admission signals** so external GPU/CPU/memory pressure can influence placement without PAIR needing to manage those workloads.
7. **Explainable routing telemetry** showing candidates, exclusions, selected node, selection reason, and failovers, without logging prompt content.
8. **Optional telemetry hardening** under the same cluster trust/mTLS model used by other PAIR services.

The observable result would be safer and more predictable scheduling across heterogeneous nodes, lower model load/unload churn and tail latency, and easier integration with higher-level orchestration systems while PAIR remains the inference transport and failover layer.

Guidottavio D’Assisti
guydax@gmail.com

### Alternatives considered

[...]

PR #23, opened 2026-09-05
Expose observed loaded-model VRAM telemetry

Ayushkumar111:
## Summary
- expose observed loaded-model GPU-resident bytes from the engine manager
- propagate telemetry through node scanning and UI broker discovery payloads
- preserve unknown versus known-zero values and document that this is telemetry, not a routing admission requirement

Related to #7. This PR intentionally does not reject or rank nodes based on size_vram; that field is observed residency and can represent CPU/GPU offload.

## Validation
- go test ./... in services/nvpair-engine-manager
- go test ./... in services/nvpair-node-scanner
- go test ./... in services/nvpair-ui-broker
- go test ./... in services/shared

### DIFF
--- a/services/nvpair-engine-manager/manifests/ollama.json
+++ b/services/nvpair-engine-manager/manifests/ollama.json
@@ -88,10 +88,10 @@
       "http": { "method": "GET", "path": "/api/tags" },
       "result": { "array": "models", "field": "name" }
     },
-    "loaded_models": {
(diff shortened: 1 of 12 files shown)


Noah-Tervalon-Nvidia:
Thanks for your interest in PAIR and for making a PR! Just adding a note here. As noted in the feature request, Sherief is working on a handful of scheduler updates. Once we get those in I think we should revisit this and decide if it's still needed (I think it will be a nice to have).

PR #27, opened 2026-09-05
Add Anthropic Messages API Routing

### LINKED ISSUE #16: [Feature]: Expose the Anthropic Messages API (POST /v1/messages) on the PAIR proxy
PAIR pools capacity so that a machine too small to hold a useful model can still use one. Coding agents that speak the Anthropic Messages API are a large and growing class of client that cannot reach PAIR at all today, and the reason is protocol rather than capability: both engines PAIR already routes to serve `POST /v1/messages` natively, and PAIR re-exposes them as OpenAI only. Adding the endpoint alongside the existing ones would open pooled local capacity to those clients without changing anything for current ones.

### Area

API compatibility

### User problem

Who is affected: people running an agentic coding client that speaks the Anthropic Messages API against local models, across a mix of machines where no single one is large enough to hold a useful model roster. That is the case PAIR is built for, and it is ours: a team where a few machines have 64 GB or more and most have far less.

What cannot be accomplished today: such a client sends every request to one base URL and speaks only `/v1/messages`. PAIR presents Ollama-compatible and OpenAI-compatible proxy endpoints, so the client cannot reach PAIR at all.

The gap is narrow rather than architectural, because the destination engines already understand the dialect:

- Ollama's v0.14.0 release notes read "Anthropic API compatibility: support for the `/v1/messages` API".
- LM Studio's post "Use your LM Studio Models in Claude Code" (2026-01-30) reads "Run Claude Code with any local model using LM Studio's Anthropic-compatible API".

Checked directly rather than taken from the notes: on LM Studio 0.4.20 and Ollama 0.33.2, both return well-formed Anthropic Messages responses, content blocks and `stop_reason` and `usage` included.

So a request arrives in a dialect the destination engine already speaks, and the only component in the path that does not is the proxy in the middle.

### Desired outcome

`POST /v1/messages` on the PAIR proxy, alongside the existing endpoints, routed by the same scheduler and answered by the selected node's engine. Observable behavior:

- A client configured with PAIR's endpoint as its Anthropic base URL completes a normal request and receives a well-formed Messages response.
- Streaming returns the Anthropic SSE event grammar rather than a translated approximation.
- `tool_use` and `tool_result` content blocks survive intact in both directions. This is the part that matters most for agent clients, because a dropped or reshaped tool block fails silently and looks like a confused model rather than a transport bug.
- `/v1/messages/count_tokens` behaves predictably. Agent clients call it, and neither engine implements it: at LM Studio 0.4.20 it returns HTTP 200 carrying `{"error":"Unexpected endpoint or method."}`, and at Ollama 0.33.2 it returns 404. Passing it through where an engine implements it and returning a clear error where none does would improve on both, and PAIR is not the right place to invent a tokenizer.
- Model selection continues to work the way it does on the existing endpoints.

Passthrough rather than translation looks like the smaller and safer change: when the selected node's engine speaks Anthropic Messages, forwarding the body preserves fidelity by construction and needs no per-field mapping to maintain as the API evolves.

### Alternatives considered

- Point the client directly at a single node's LM Studio or Ollama. This works, it is what we do today, and it gives up everything PAIR exists to provide: no pooling, no scheduling, no use of idle machines.
- Put a translating gateway in front of PAIR. This reintroduces Anthropic to OpenAI conversion at exactly the point where the destination engine did not need it, and tool-call round trips are the known casualty of that conversion.
- Run a small Anthropic-to-OpenAI shim alongside PAIR. Same fidelity problem, plus another process on every machine.

Each alternative pays a translation cost to reach engines that already speak the dialect.

### Compatibility and security implications

Additive. Existing Ollama-compatible and OpenAI-compatible endpoints are unchanged, and clients that do not use `/v1/messages` see no difference. No change to discovery, pairing, or the trust model is implied.

Routing an unfamiliar path to a node whose engine does not implement it should fail with a clear error rather than a partial or reshaped response, so that a capability gap stays visible instead of silent.

### Validation approach

- A three-call check against each supported engine: an ordinary message, a message that must produce a `tool_use` block with valid arguments, and a streaming request whose SSE event sequence is compared against the Anthropic grammar.
- A `count_tokens` call that returns either a clear answer or a clear error, never a hang and never an HTTP 200 carrying an error body.
- The same three checks with the request routed to a remote node rather than the local one, confirming the shape survives the hop.

### Scope of what I checked

I read the proxy service only, not every service, and I did not locate in source the Ollama-native surface the README describes. The engine behavior above was executed at the pinned versions named; the claim that PAIR has no Anthropic surface today comes from finding no `/v1/messages` or Anthropic reference in the repository.

This sits next to two open threads rather than pointing in a fresh direction: #9 adds vLLM as a third engine, which serves the same endpoint, and #7 asks for richer placement signals, which is the same question of whether a node can serve a given request rather than merely being idle.

Happy to test on Apple Silicon and report back, and happy to follow up with a pull request if this is a direction you would take.

### Confirmations

Ratherrayees:
## Summary

Include support for routing requests to the Anthropic Messages API through the PAIR Ollama and LM Studio proxies.

The POST /v1/messages endpoint has now been identified as an inference endpoint and can make use of PAIR's present model selection, routing, and failover features without needing to translate the Anthropic request into another protocol.

Compatibility with Anthropic API clients such as Claude Code is improved since /v1/messages requests can be passed through PAIR to compatible local inference engines, enabling users to make use of Anthropic-compatible clients while taking advantage of PAIR's existing model selection, routing, and failover features.

Fixes [#16](https://github.com/NVIDIA/Personal-AI-Router/issues/16)

## Changes

Include the POST request to the /v1/messages endpoint in the list of inference endpoints for the Ollama and LM Studio proxies.
Include test coverage to verify model-based candidate selection and request-body passthrough for Anthropic Messages requests.
Update the proxy and the getting-started documentation.
* Bump service versions:

  * `ollama-proxy`: `0.27.0`
  * `lmstudio-proxy`: `0.17.0`

## Design

The implementation employs passthrough rather than protocol translation; since the available engines already support the Anthropic Messages API, PAIR sends the request to the chosen engine and keeps the engine's original Anthropic response format.

The alteration only makes it possible to access `/v1/messages` via the inference routing pathway that PAIR already has in place, and the existing Ollama-native and OpenAI-compatible endpoints stay exactly the same.

### Automated

Checked the proxy routing behaviour for '/v1/messages'.
The candidate selection and request-body passthrough have been verified through the routing tests.
Carried out the cross-process service test suite using the command `cd services/tests && go test ./...`.

### End-to-end

We validated the POST request to /v1/messages via the PAIR proxy using Ollama and qwen3:4b.
Confirmed that the request had been routed via PAIR and had produced an Anthropic-shaped Messages response.
Confirmed that the current OpenAI-compatible endpoint is still working correctly.

## Notes

The implementation is intentionally directed at making the Anthropic Messages endpoint usable with PAIR's current routing and failover facilities and it does not include Anthropic-specific request/response translation or provide a separate routing path for Anthropic.


Noah-Tervalon-Nvidia:
Glad to have you here contributing to the project! This lgtm - I am currently working on some major redesigns for the proxies to unify them. I want to wait on merging in changes for the proxies until we get that out to avoid extra work in merging into the new version so I expect this will need a minor rework once we get that out and we'll wait on merging it until then.

Ratherrayees:
Thanks for taking the to review it and heads-up on the proxy rework. Sounds good - I’ll hold off until the proxies are redesigned and then make the necessary adjustments once it’s ready. I’m looking forward to being able to contribute more to the project!

kjlubick:
Notes: https://lmstudio.ai/docs/developer/anthropic-compat and https://docs.ollama.com/integrations/claude-code are where the currently supported engines document their Anthropic pass-through support (which this PR adds).

services/lmstudio-proxy/failover_test.go
```diff
@@ -598,3 +598,83 @@ func TestResolveCandidates_SelfGuard(t *testing.T) {
 		t.Errorf("expected the real node to survive the self-guard, candidates = %+v", cands)
 	}
 }
+
+// TestHandleHTTP_AnthropicMessages_InferenceRouting proves /v1/messages is
+// treated as an inference endpoint: model-based candidate filtering is applied
+// and the request body is forwarded unchanged to the matching node.
+func TestHandleHTTP_AnthropicMessages_InferenceRouting(t *testing.T) {
```
kjlubick: Thanks for adding this test. I'm going to upload a follow-on commit that reduces the duplication here with a table-based test (that seemed easier than me describing what I had in mind).

kjlubick:
One more review comment, could you add something in the description about improving support for Claude Code (I'm guessing that's the motivation) or whatever problem you bumped into that inspired this change? It's nice to include the "how does this benefit the user" a bit more explicitly in changes. Thanks again for your contribution!

Ratherrayees:
Thanks Kaylee! I’ve improved the PR description to better describe the user benefit and Claude Code compatibility. Also really appreciate you adding the test refactor based on tables, it's a much cleaner way to express the test coverage.

Ratherrayees:
I really enjoy working with PAIR and look forward to doing more for the project.

Fingers crossed it gets merged!

kjlubick (approved):

kjlubick:
Just to touch base here, we are working on getting our final batch of internal changes out into the public repo. Once those are in, we'll get this rebased (if necessary) and landed. Thanks again for your contribution!

Noah-Tervalon-Nvidia:

Ratherrayees:
Hi Noah, thanks again for the kind words and for the heads-up about the proxy redesign.

I’ve now rebased the PR onto the current `develop` and ported the Anthropic Messages API support to the new unified `nvpair-proxy` architecture.

The implementation now:

* Adds `POST /v1/messages` to the shared inference route set used by both Ollama and LM Studio facades.
* Keeps the existing `ollama-proxy:` and `lmstudio-proxy:` client-facing namespaces unchanged.
* Reuses the existing model selection, eligibility, reservation, retry, and failover flow without introducing Anthropic-specific routing or protocol translation.
* Updates the unified proxy tests, cross-process strict model-routing coverage, and README.
* Incorporates the table-based test structure from the earlier review.

For validation:

* Unified `nvpair-proxy` tests pass.
* Anthropic routing/failover cases pass for both Ollama and LM Studio.
* The full `services/tests` suite has one failure in `TestBrokerProxySetPortRebinds`; I reproduced the same failure on a clean current `origin/develop` baseline with the same port/settings resolution error.
* SPDX and `git diff --check` pass.
* I also verified an end-to-end Claude Code session through PAIR using Ollama.

The updated branch is pushed at `7749b54`. Thanks again for pointing me in the right direction with the redesign.

kjlubick:
You and I were apparently trying to rebase at the same time. You got in first, I pushed a few other changes on top of yours (e.g. documentation and cleanup tests). I think we should be good to go

Ratherrayees:
Hehe, looks like we were racing each other there 😄 Glad we got it sorted. Thanks for jumping in and cleaning things up!

services/nvpair-proxy/engines.go
```diff
@@ -116,43 +116,53 @@ type engineProfile struct {
...
+var anthropicInferenceRoutes = []route{
+	{Path: "/v1/messages", Role: roleInferencePOST},
+}
+
 var profiles = buildProfiles()
 
 func buildProfiles() []engineProfile {
 	ollama, _ := engines.ByName("ollama")
 	lmstudio, _ := engines.ByName("lmstudio")
+	ollamaRoutes := []route{
```
Noah-Tervalon-Nvidia, line 136: super minor nit, but could we move this list of ollamaRoutes to a variable similar to the anthropic or openai routes and then build up the list again inside this function with something like:
`ollamaRoutes := append(ollamaInferenceRoutes, openAIInferenceRoutes...)`
`ollamaRoutes = append(ollamaRoutes, anthropicInferenceRoutes...)`

just thinking about future functionality, perhaps there exists a world where someone else adapts ollama's inference routes and we would want to just take that from the base list. Would do the same thing for lmsroutes

kjlubick: Done.

Noah-Tervalon-Nvidia (approved):

Issue #22, opened 2026-09-05
[Task]: No public CI workflows, and the Makefile references a ci/pipeline.yml that is not in the repository


imluffy000:
### Task type

Other

### Current state

The repository has no CI configuration. At commit 13b6811 there is no
.github/workflows/ directory and no ci/ directory; .github/ contains only
ISSUE_TEMPLATE/ and PULL_REQUEST_TEMPLATE.md.

The Makefile refers to a pipeline file that is not in the tree. Makefile:41-44:

    # Mirrors the repository-wide header gate plus the desktop half of the CI
    # validate and monorepo-gate stages (ci/pipeline.yml). CI additionally rebuilds
    # the binaries and runs the Go suites; use make build-binaries and
    # make test-services for those.

ci/pipeline.yml does not exist here, so a contributor reading the Makefile is
pointed at something they cannot open, and cannot tell which gates a maintainer
will actually run against their pull request.

The consequence for an outside contributor: CONTRIBUTING.md asks them to run the
gates by hand, nothing verifies a pull request automatically, and a missing SPDX
header or a failing Go suite surfaces only in human review - which CONTRIBUTING.md
says will send the change back.

### Desired result

Concrete completion condition - either of the following.

(a) A GitHub Actions workflow that runs on pull_request and on push to main and
    executes the documented gates, so a contributor sees the same verdict a
    maintainer would:

      - node scripts/spdx-headers.mjs   (dependency-free; runs in a fresh clone
        before any install)
      - the desktop gates: npm ci, npm run verify:build-scripts, npm run lint,
        npm run typecheck, npm run service-contracts:check, npm run test:unit,
        npm run dead-code:check
      - go test ./... in every services module, plus services/tests

    "make check" (Makefile:45) and "make test" (Makefile:47) already wrap most of
    this on Linux and macOS; note that "make check" deliberately omits
    dead-code:check, per AGENTS.md.

    A matrix over ubuntu-latest / macos-latest / windows-latest would match the
    stated support surface. A Linux-only job would still be a large improvement
    over none.

(b) If CI is intentionally internal-only and not mirrored to this repository,
    then correct Makefile:42 so it does not reference a path that is absent here,
    and say in CONTRIBUTING.md that pull requests are validated by an internal
    pipeline that contributors cannot see or run.

Either one resolves the mismatch. (a) additionally gives contributors a
pre-review signal.

### Relevant paths or references

- Makefile:42 - the ci/pipeline.yml reference
[...]

PR #31, opened 2026-09-05
docs: run PAIR in a container or on Kubernetes, with a Dockerfile and manifests

DustinTrap:
## Description

PAIR has no story for running a node as a container or as a Kubernetes pod, and two properties of the design make the obvious attempt fail silently: mDNS never leaves the local link, so a container on a pod network discovers nothing; and the endpoint is loopback-only, so every other workload on the cluster gets `403 loopback-only`.

This adds `docs/containers.mdx`, which explains both and what to do about them, and `services/container/` with a Dockerfile that builds a services-only node from the release asset (pinned by sha256, UBI 9 minimal base, arbitrary user id), an entrypoint, a dependency-free loopback relay with a source allowlist, and Kubernetes manifests (StatefulSet with a persistent volume for the node's identity, Service). The page is registered in the Fern navigation after Building and Running, and `building.mdx` and `troubleshooting.mdx` link to it from the two places a reader arrives with this problem.

User-visible outcome: a workload on a Kubernetes cluster can call `http://pair.pair.svc.cluster.local:1234/v1` and have PAIR route it to a node with an engine.

Alignment: #32 proposes this distribution channel. Related: #30 is the address-ranking fix for OVN-Kubernetes hosts; the page mentions the symptom and the manual-add workaround. Scripted pairing (#11) would remove the attach step.

## Scope

In: the documentation page, the Dockerfile and entrypoint, the relay, the manifests, navigation and cross-links.

Out: building the image in CI (the repository has no public CI, #22); publishing an image anywhere; a Go relay (the Python one needs nothing built, and can be ported if maintainers prefer a fourteenth binary); multi-node-per-host layouts; any change to the services.

## Validation

Environment: PAIR v0.1.1. Single-node OpenShift 4.22 (OVN-Kubernetes) as the container host, clustered with a Linux node holding one NVIDIA GPU (LM Studio engine) and a macOS node.

- Image built with the Dockerfile; smoke-tested as an arbitrary UID: thirteen binaries present and executable, data root writable, relay starts and accepts, `nvpair-ui-broker` answers a JSON-RPC `ping` (on a native x86_64 runner).
- Pod admitted under the `hostnetwork-v2` SCC with seccomp `RuntimeDefault`, all capabilities dropped, no privilege escalation, arbitrary UID. Pod IP is the node's LAN address; the peer's mTLS ports 14318-14323 reachable from the pod; mDNS multicast heard.
- From another namespace: `http://<node-ip>:1234` returns `403 loopback-only`; `http://pair.pair.svc:1234` returns `503 model inventory unavailable` before pairing, a completion after.
- Paired by invite from the Linux node and PIN entry through `kubectl attach`. Restarted the pod; identity persisted on the PVC and the member stayed listed.
- Five-minute soak from an in-cluster Job: 71 requests, 71 ok, mean latency 4.3 s, about 76 tokens/s, GPU about 95% busy, total LAN traffic a few megabytes. Model: a 27B dense model at 64k context, one request at a time, warm.
- `node scripts/spdx-headers.mjs`: 0 missing.

## Risk

The relay has no authentication and is the trust boundary; it enforces that boundary itself by refusing any source outside its allowlist (loopback and the pod network), because NetworkPolicy does not apply to host-network pods. No Ingress or Route. The page says so. (Corrected in 00b0ba6: the first revision shipped a NetworkPolicy that could not apply.) The image runs `nvpair-tui` in the foreground, so a crashed node restarts the pod rather than lingering. The SCC binding is OpenShift-specific and inert elsewhere (the ClusterRole does not exist), noted in the page. Nothing in `services/` changes and no binary is rebuilt, so no version bump.

## Checklist

- [x] New or existing tests cover the change. (Validation above; no Go or TypeScript changes.)
- [x] I bumped any affected component in `services/versions.json`, and described user-visible changes above so they reach the release notes. (No compiled output changed; no bump.)

### DIFF
--- a/docs/building.mdx
+++ b/docs/building.mdx
@@ -163,6 +163,14 @@ compete for the same services, engines, and ports.
 after it opens: pairing, engines, models, routing, and settings. It also lists
 the operations that remain desktop-only.
 
+### Run in a Container
+
+The services bundle needs nothing from the host but a terminal and a writable
+data directory, so it also runs as a container, and as a pod on Kubernetes.
+[Containers and Kubernetes](containers.mdx) has a Dockerfile, the manifests, and
+the two things a container must get right: host networking for discovery and a
+loopback relay for the endpoint.
+
 ### Run the Broker Yourself
 
 Run `nvpair-ui-broker` directly when you want to drive the services
--- a/docs/containers.mdx
+++ b/docs/containers.mdx
@@ -1,0 +1,173 @@
+{/*
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+*/}
+
+# Running PAIR in a Container or on Kubernetes
+
+The services-only build of PAIR is thirteen statically linked Go binaries with
+no cgo, so a container image is a base plus an unzip. This page covers what a
+container must get right, gives a Dockerfile that builds a node from a release
+asset, and gives Kubernetes manifests that run it as a pod and expose its
+endpoint to other workloads on the cluster.
+
+The node routes; it does not infer. It joins your existing cluster and forwards
+requests to the nodes that have engines, so a workload on Kubernetes reaches
+the GPUs on your other machines.
+
+Everything here was validated with PAIR v0.1.1 on a single-node OpenShift 4.22
+cluster running OVN-Kubernetes, clustered with a Linux node holding an NVIDIA
+GPU and a macOS node. The manifests are plain Kubernetes except for one
+OpenShift-specific role binding, called out below.
+
+## What a Container Must Get Right
+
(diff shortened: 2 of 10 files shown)


DustinTrap:
Pushed 00b0ba6, a correction to the first revision: the NetworkPolicy it shipped could not apply, because NetworkPolicy does not govern host-network pods, and the relay listened on `0.0.0.0`, which on a host-network pod is the LAN. Verified on OVN-Kubernetes that the relay answered an unauthenticated request from a LAN host. The relay now refuses any source outside `--allow-cidr` (loopback and `10.128.0.0/14` by default) with a canned 403, the policy is gone from the manifests, and the page and the limits list say why. #38 proposes the native, authenticated alternative; if it lands, these manifests can point the Service at the proxy port with a key and drop the relay.

PR #34, opened 2026-09-05
fix(ui): show every GPU on a node card, not just the first

### LINKED ISSUE #33: [Bug]: Node card header shows only the first GPU, so multi-GPU hosts are misreported as single-GPU
### PAIR version or commit

0.1.1 (`13b68115fa2c9c1d94f1ead1358f8d5a527cfecf`), installed from `NVPAIR-Setup-0.1.1-arm64.dmg` and reproduced against a source build of the same commit.

### Affected component

Desktop application

### Environment

```
OS:               macOS 26.6.2 (viewing node), Rocky Linux 9.8 (affected node)
Architecture:     arm64 (viewer), x86_64 (affected node)
GPU and driver:   affected node has TWO GPUs -
                  Tesla P100-PCIE-16GB + NVIDIA GeForce RTX 3060,
                  driver 580.159.04, CUDA 13.0
Engine and version: Ollama (PAIR-managed)
Model:            qwen3:1.7b
Cluster size:     4 nodes
```

### Steps to reproduce

1. Run PAIR on a host with **two or more GPUs**.
2. Pair it into a cluster and open the desktop application's **Overview**.
3. Look at that node's card header.
4. For comparison, read the same node's raw inventory: `curl http://<node>:14318/v1/node-info`.

### Expected behavior

The node card header identifies the host's GPUs. A machine with a P100 *and* a 3060 should not be indistinguishable from a machine with only a 3060.

### Actual behavior

The header shows exactly **one** GPU name. The second card is absent, with no `+1`, no count, and no other indication that more GPUs exist — so a multi-GPU host is silently misreported as single-GPU in the summary view.

The data is correct at every other layer:

- `/v1/node-info` returns **both** GPUs with correct VRAM totals.
- The detailed performance view charts **both**, per its own comment: *"The detailed performance view intentionally charts EVERY detected GPU."*
- `inferenceHardwareIds` is **absent** from the payload, which by the contract in `src/ui/utils/gpu-inference.ts` means "show every GPU" — so nothing is being filtered as un-ready.

Only the header label drops them.

**Cause** — `desktop/src/ui/components/NodeList/NodeCardDetails.tsx:276` passes only the first element:

```tsx
gpuLabel={gpuInfo.length > 0 ? gpuInfo[0].name : undefined}
```

`NodeLabel` accepts a single `gpuLabel: string | undefined` and pushes one `gpu` segment, so every GPU after the first is unrepresentable by construction.

Worth noting the fix is not simply joining the names into one string: `NodeLabel` applies wrapping, `·` separators and truncation **per segment**, so a joined string is a single segment that truncates after the first name on a narrow card — the same blind spot in a different form. One segment per GPU preserves the existing layout behavior.

This matters more on larger hosts: an 8-GPU server currently presents as a 1-GPU server in the node list.

```shell
# /v1/node-info from the affected node - both GPUs present and correct.
# Host/cluster identifiers and addresses removed per the template.
{
  "GPUs": [
    { "name": "Tesla P100-PCIE-16GB",    "vram_bytes": 17179869184, "vram_used_bytes": 16077815808 },
    { "name": "NVIDIA GeForce RTX 3060", "vram_bytes": 12884901888, "vram_used_bytes": 12120489984 }
  ],
  "cpu": { "name": "AMD Ryzen 5 5600X 6-Core Processor", "cores": 6, "utilization_percent": 1 },
  "memory": { "total_bytes": 139586437120, "used_bytes": 14764740608 },
  "telemetryValid": true
}

# Desktop node card header for that same node:
#   <NODE> - <address> - NVIDIA GeForce RTX 3060
# The Tesla P100 does not appear.
```

### Confirmations

- [x] I searched existing issues for duplicates. The closest is #4 (local node's GPU/CPU/memory enrichment never refreshed), which concerns *stale* discovery data and explicitly notes the UI is unaffected; this is a rendering defect with fresh, correct data.

CryptoJones:
Closes #33.

## Problem

`NodeCardDetails.tsx:276` passed only the first GPU to the node card header:

```tsx
gpuLabel={gpuInfo.length > 0 ? gpuInfo[0].name : undefined}
```

`NodeLabel` took a single `gpuLabel: string | undefined`, so every GPU after the first was unrepresentable by construction. A host with a Tesla P100 **and** an RTX 3060 rendered identically to one with only the 3060, with no count or overflow hint that anything was missing.

The data was never wrong — `/v1/node-info` reports every GPU, `inferenceHardwareIds` was absent (which by the contract in `gpu-inference.ts` means "show every GPU"), and the detailed performance view already charts all of them. Only the header dropped them.

## Change

Emit **one label segment per GPU**.

Segments are the right unit rather than a joined string: `NodeLabel` applies wrapping, `·` separators and truncation *per segment*, so `gpus.join(' · ')` would be a single segment that truncates after the first name on a narrow card — the same blind spot in a new shape. Per-GPU segments keep the existing layout behavior and let each name wrap and truncate on its own.

Three supporting details:

- The segment builder moves to `ui/utils/node-label-segments.ts`. The unit project runs `environment: 'node'` with no DOM and deliberately excludes components, so a pure helper is testable there without adding jsdom/testing-library to the tree. It also lands inside the existing coverage scope ("the renderer's helper and constant modules").
- `gpuLabels` is memoized in `NodeCardDetails`. Building the array inline gave it a new identity every render, and `NodeLabel` observes its segments — that would have torn down and rebuilt the `ResizeObserver` on every render.
- Blank GPU names are skipped, so a backend reporting an empty name cannot render a separator with nothing after it.

Behavior for 0- and 1-GPU hosts is unchanged.

## Tests

`desktop/tests/modular/node-label-segments.test.ts` — 13 cases:

- name/address ordering, and the address-as-primary fallback
- single-GPU host unchanged
- the reported dual-GPU regression, asserting topology order
- `it.each([3, 4, 5, 6, 7, 8])` — every GPU present, in order, at each count; 8-GPU servers are the case that matters most here
- unique, stable `gpu-N` ids across an 8-GPU host so React keys do not collide
- blank names dropped without disturbing the rest
- no GPU segments for a host reporting none

## Verification

Run from `desktop/`, all green on this branch:

| Gate | Result |
|---|---|
| `npm run lint` | exit 0 (one pre-existing warning in `node-info-poller.ts`, untouched here) |
| `npm run typecheck` | exit 0 |
| `npm run dead-code:check` | exit 0 — no dead code |
| `npm run test:unit` | 38 files, 221 tests passed |

---

*Proudly Made in Nebraska. Go Big Red! 🌽 <https://xkcd.com/2347/>*

### DIFF
--- a/desktop/src/ui/components/NodeList/NodeCardDetails.tsx
+++ b/desktop/src/ui/components/NodeList/NodeCardDetails.tsx
@@ -100,6 +100,11 @@ function NodeCardDetails({ node }: NodeCardDetailsProps) {
         })
     }, [hasGpu, node, nodeMetrics])
 
+    // Memoized so the label's segment list keeps a stable identity across
+    // renders; NodeLabel observes its segments and would otherwise rebuild the
+    // ResizeObserver every render.
+    const gpuLabels = useMemo(() => gpuInfo.map(gpu => gpu.name), [gpuInfo])
+
     const cpuFallbackInfo: CpuFallbackInfo | null = useMemo(() => {
         if (hasGpu) return null
         const ramTotal = node.topology.ram
@@ -273,7 +278,7 @@ function NodeCardDetails({ node }: NodeCardDetailsProps) {
                             <NodeLabel
                                 name={node.name}
                                 ipAddress={node.ipAddress}
-                                gpuLabel={gpuInfo.length > 0 ? gpuInfo[0].name : undefined}
+                                gpuLabels={gpuLabels}
                                 isLocal={isLocal}
                             />
                         </div>
--- a/desktop/src/ui/components/NodeList/NodeLabel.tsx
+++ b/desktop/src/ui/components/NodeList/NodeLabel.tsx
@@ -4,37 +4,26 @@
 import { Flex, Text } from '@nvidia/foundations-react-core'
 import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react'
 import { LocalBadge } from '@/ui/components/LocalBadge'
-
-interface LabelSegment {
-    id: string
-    text: string
-}
+import { buildNodeLabelSegments } from '@/ui/utils/node-label-segments'
 
 export default function NodeLabel({
     name,
     ipAddress,
-    gpuLabel,
+    gpuLabels,
     isLocal
 }: {
     name: string
     ipAddress: string
-    gpuLabel: string | undefined
+    gpuLabels: string[]
     isLocal?: boolean
 }) {
     const segmentRefs = useRef<(HTMLDivElement | null)[]>([])
     const containerRef = useRef<HTMLDivElement | null>(null)
     const [separatorAfter, setSeparatorAfter] = useState<boolean[]>([])
-    const segments = useMemo(() => {
-        const primaryText = (name || ipAddress).toUpperCase()
-        const nextSegments: LabelSegment[] = [{ id: 'primary', text: primaryText }]
-        if (name && ipAddress) {
-            nextSegments.push({ id: 'ip', text: ipAddress })
-        }
-        if (gpuLabel) {
-            nextSegments.push({ id: 'gpu', text: gpuLabel })
-        }
-        return nextSegments
-    }, [gpuLabel, ipAddress, name])
+    const segments = useMemo(
+        () => buildNodeLabelSegments(name, ipAddress, gpuLabels),
+        [gpuLabels, ipAddress, name]
+    )
 
     const updateSeparators = useCallback(() => {
         const next = segments.map((_segment, index) => {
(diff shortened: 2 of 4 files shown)


CryptoJones:
Verified on real hardware — the desktop app running this branch, on a 4-node cluster where one host has two GPUs:

PAIR node list showing a dual-GPU host with both cards named

`PLUTO.LOCAL.TLD` now reads **`NVIDIA GeForce RTX 3060 · Tesla P100-PCIE-16GB`**. On `main` that same row shows only `NVIDIA GeForce RTX 3060` — the Tesla P100 is absent.

The other three nodes are single-GPU and render exactly as before, which is the no-regression half of the check.

LAN addresses are redacted per the issue template's guidance on private network details. Image is hosted on an orphan branch of the fork so it stays out of this PR's diff.

sherief-nv:
Thank you for the contribution! The team is evaluating this for inclusion in the next release. I'm changing the target branch to the development branch, which we will eventually merge with main when a release is ready.

Noah-Tervalon-Nvidia:
@ckelseynv to give final stamp on UI changes. This lgtm.

PR #37, opened 2026-09-06
Fix HTTP connection reuse in engine health probes

Psych0h3ad:
## Description

Periodic engine health checks close HTTP response bodies without reading them. With Go 1.26.7, this prevents the HTTP/1 transport from reusing those connections and opens a new connection for every poll. Draining ordinary responses before closing reduces unnecessary connection churn while PAIR monitors Ollama and LM Studio.

The regression tests reproduce 32 connections for 32 polls before the fix and one connection for the same 32 polls after it, for both Content-Length and chunked responses across all three health-check paths.

## Scope

- Drain response bodies before closing in the broker's Ollama and LM Studio checks and the engine manager's HTTP probe, including unsuccessful HTTP statuses.
- Limit each drain to 1 MiB and retain the existing broker client timeout of 2 seconds and engine-manager probe deadline of 3 seconds.
- Add tests for HTTP/1 connection reuse, unchanged status-based health decisions, bounded reads, and stalled-body deadlines.
- Bump `nvpair-ui-broker` from 0.40.2 to 0.40.3 and `nvpair-engine-manager` from 0.17.4 to 0.17.5.

No API, configuration, dependency, or desktop changes are required. This change addresses connection churn; it does not reset existing TCP state or establish a cause for any operating-system TCP reclamation failure.

## Validation

Targeted tests passed on macOS arm64 with Go 1.26.7 and `CGO_ENABLED=0`. In the commands below, `go` is the Go 1.26.7 executable; its machine-specific installation path is omitted.

From `services/nvpair-ui-broker`:

```sh
CGO_ENABLED=0 GOPROXY=off GOTOOLCHAIN=local go test -count=1 -run '^TestHealthChecks' -v .
```

From `services/nvpair-engine-manager`:

```sh
CGO_ENABLED=0 GOPROXY=off GOTOOLCHAIN=local go test -count=1 -run '^TestProbeHTTP' -v .
```

The connection-reuse cases use the real `net/http` HTTP/1 client and server over `net.Pipe`, requiring no available TCP source ports. Each case alternates successful and unsuccessful statuses over 32 sequential requests. All six combinations of health-check path and response framing failed against the original production code with 32 connections and passed after the fix with one connection. These are transport regression results, not measurements of kernel TIME_WAIT recovery.

The complete package suites have not been verified: existing engine-manager TCP tests encountered source-port exhaustion in the test environment, and the broker suite encountered an existing macOS configuration-isolation defect. These limitations are separate from the changed health checks. Full desktop checks were not run because the patch changes only these two Go services. Live engine operation has not been verified.

`node scripts/spdx-headers.mjs` passed across the repository (873 checked, no missing or unclassified headers). `git diff --check` passed for the change. Both modified service binaries built successfully with the same Go toolchain.

## Risk

Health results continue to depend on the HTTP status, including when a body read fails. A stalled response can now keep the check active until the existing timeout or deadline, and an oversized or incomplete body can still cause the connection to be discarded. The drain is bounded in size and time and discards content without logging it. No data migration or wire-format change is involved.

## Checklist

- [x] Relevant documentation is updated. No documentation changes are needed for this internal connection-lifecycle fix; the user-visible behavior is described above for release notes.


Noah-Tervalon-Nvidia:
Thank you for bringing this issue up. We're going to have a look and validate that everything looks good on it.

kjlubick:
I think this is worth fixing, although I noticed on golang 1.27, the tests w/o the fix only had 2 or 3 open connections. This seems to be due to `HTTP/1 Response.Body now automatically drains any unread content upon being closed, up to a conservative limit, to allow better connection reuse. For most programs, this change should be a no-op, or result in a performance improvement` (https://go.dev/doc/go1.27).

I'll send a commit to the branch to make the change be a bit more succinct and align with internal standards.

kjlubick (approved):

kjlubick:
Just to touch base here, we are working on getting our final batch of internal changes out into the public repo. Once those are in, we'll get this rebased (if necessary) and landed. Thanks again for your contribution!

Psych0h3ad:
> Just to touch base here, we are working on getting our final batch of internal changes out into the public repo. Once those are in, we'll get this rebased (if necessary) and landed. Thanks again for your contribution!

Yay! I was struggling with this when I installed for my personal AI bot brain Mac Mini😁

kjlubick:
We'll need to rebase this onto develop (which has had some major changes lately). I can try to do that this morning.

kjlubick (approved):

Noah-Tervalon-Nvidia (approved):

PR #40, opened 2026-09-06
fix(desktop): swallow EPIPE from structured logger stdout

smfworks:
## Outcome

A closed `stdout` pipe no longer takes down the Electron main process with
`write EPIPE` / “A JavaScript error occurred in the main process”.

## Bug

`writeEntry` already treats log-file append as best-effort. Console output
called `process.stdout.write` with no handler. When PAIR is launched from a
parent that later closes the pipe (or after a long sleep with a dead stdout),
Node emits `write EPIPE` from `afterWriteDispatched`. That is an
`uncaughtException`; `electron-log-unhandled` shows the dialog.

Reproduced on Linux 0.1.1: second launch against a singleton that inherited a
dead stdout.

## Fix

- `try/catch` around `process.stdout.write` (same shape as the file write).
- Attach no-op `'error'` listeners on `stdout`/`stderr` in `initFileLogger`
  so the *async* EPIPE event is not uncaught. `try/catch` alone does not
  cover that path.

## Scope

In: `desktop/src/shared/utils/log.ts` and a unit test.
Out: launcher wrapping, asar packing, service binaries. No
`services/versions.json` bump (desktop JS only).

## Validation

- Environment: Linux, Node 26.7, `npm run test:unit -- tests/modular/structured-log-stdout.test.ts` — 2 passed.
- `npm run typecheck` — pass.
- `node scripts/spdx-headers.mjs` — 0 missing.

## Compatibility / security

Logging to a closed stdio is dropped, same as a failed file append. No change
to what is logged. No new dependencies.

## Documentation

None. Logger contract is unchanged aside from not crashing the process.


Noah-Tervalon-Nvidia:
Thanks for raising this, seems like a miss on our part. @ckelseynv will have a look!

Issue #26, opened 2026-09-05
[Feature]: Allow manifest actions to send HTTP headers, so engines behind authentication can be managed


jlacroix82:
### Area

Engine or model management

### User problem

`ActionHTTP` in the engine manifest schema carries only a method, a path and a body schema:

```go
type ActionHTTP struct {
    Method     string          `json:"method"`
    Path       string          `json:"path"`
    BodySchema json.RawMessage `json:"body_schema,omitempty"`
}
```

There is no way to send a request header, so any engine whose API requires authentication cannot be driven by a manifest at all. This is not an edge case — an engine that binds a loopback port and expects a bearer token is a normal shape, and it is the shape you get whenever a local engine also supports remote access.

The concrete example I ran into is Unsloth Studio. Its `/api/health` is open, so PAIR's readiness and health probes work fine and the engine looks perfectly manageable. But `/v1/models` returns:

```
HTTP 401  {"error":{"message":"Not authenticated","type":"authentication_error"}}
```

so `list_models`, `loaded_models` and every model operation fail. The engine can be started and monitored but not used, which is arguably worse than not supporting it at all, because the failure appears only after it looks healthy.

The same applies to vLLM started with `--api-key`, to SGLang behind a token, and to any engine reached through an authenticating proxy.

### Desired outcome

An optional `headers` map on `ActionHTTP`, so a manifest can declare what an action needs to send:

```json
"list_models": {
  "http": {
    "method": "GET",
    "path": "/v1/models",
    "headers": { "Authorization": "Bearer {api_key}" }
  },
  "result": { "array": "data", "field": "id" }
}
```

That implies a way to supply the secret, which is the part worth designing rather than assuming. Some options, roughly in order of how much they widen the trust surface:

- a new `{api_key}`-style placeholder resolved from the existing per-engine settings the user already edits in the app, alongside the persisted port override
- resolution from the engine's environment, reusing `runtime.env` so the value lives where the engine's other configuration already lives
- a literal string in the manifest, which is simplest but puts a credential in a file users copy around, so probably worth refusing outright

### Alternatives considered

An adapter process that holds the credential and presents an unauthenticated loopback surface to PAIR. This works — it is what I did — but it means every authenticated engine needs a sidecar to be manageable, and the credential ends up in a second place rather than in PAIR's own settings.

Skipping affected engines entirely is the current behavior, and it fails late rather than clearly: health probes pass, so the engine presents as healthy while every model action returns 401.

[...]

Issue #24, opened 2026-09-05
[Feature]: Adopt arbitrary OpenAI-compatible backends with endpoint manifests, model aliases, and deterministic capability routing


darkmatter2222:
### Area

Desktop experience

### User problem

Area

Engine or model management

User problem

PAIR solves much of the difficult infrastructure required for distributed local inference: node discovery, trust, model inventory, routing, workload visibility, and failover.

The remaining limitation for heterogeneous performance-oriented clusters is that inference backends still need to be explicitly supported as PAIR engines.

In practice, different hardware often performs best with different inference runtimes and configurations. Requiring an operator to replace an already optimized backend with a PAIR-supported engine can negate much of the reason for building a heterogeneous cluster.

I have a working three-node example:

- RTX 5090 — vLLM, Qwen3.8-27B NVFP4, primary/high-throughput node
- RTX 3090 — llama.cpp, Qwen3.8-27B Q4_K_M GGUF, secondary node
- DGX Spark / GB10 — large MoE / VLM endpoint, overflow and vision-specialist node

The router currently used for this cluster is open source here:

https://github.com/darkmatter2222/HomeAILab/tree/main/router

Its declarative endpoint configuration is here:

https://github.com/darkmatter2222/HomeAILab/blob/main/router/config/router.yaml

And the request classification / deterministic selection implementation is here:

https://github.com/darkmatter2222/HomeAILab/blob/main/router/internal/routing/classify.go

This is not intended as a competing implementation or a request for PAIR to copy this router. It is a working reference showing the behaviors that currently prevent this deployment from moving entirely to PAIR.

The core issue is broader than adding one more runtime.

PR #9 adds vLLM as another first-class engine, which is useful. But vLLM, SGLang, llama.cpp server, TensorRT-LLM gateways, custom inference servers, and future engines increasingly expose substantially the same OpenAI-compatible inference surface.

Ideally, every compatible runtime should not require a bespoke PAIR engine implementation before it can participate in the cluster.

There is also a heterogeneous routing problem.

Two nodes may expose the same logical model while having radically different:

- hardware
- quantization
- inference runtime
- context capacity
- concurrency
- prefill performance
- generation performance
- modality support
- cold-start / TTFT behavior

A request should therefore be matched against endpoint capabilities before ordinary load-based scheduling occurs.

Real-world reference behavior

The router above currently models endpoints declaratively.

For example, the same logical coding workload can be served by:

RTX 5090
vLLM
[...]

PR #50, opened 2026-09-08
Add SGLang as a fourth inference engine

cguldogan:
## Description

PAIR routes to Ollama, LM Studio and (with #9) vLLM. This adds **SGLang** as a fourth engine on the same footing: detected or adopted, started and stopped, advertised over discovery, routed to, scheduled, and shown in the desktop and terminal interfaces.

User-visible outcome:

- SGLang appears in Engine settings on Linux with install, start, stop, restart, uninstall and port configuration. On macOS and Windows it is listed but not installable, because SGLang publishes no GPU build there.
- An already running SGLang on port `30000` (a `sglang serve --model-path …` or the `lmsysorg/sglang` container) is **adopted** rather than restarted. That is the normal path.
- Like vLLM, an SGLang process serves one model chosen before it starts, so the existing **Model to serve** field applies. It accepts a Hugging Face id or a local model directory; the id SGLang then advertises is its `--served-model-name`, which defaults to the `--model-path` verbatim, so a local directory shows up under its own path. There are no pull, load, eject or delete operations.
- SGLang shares the existing OpenAI-compatible proxy with LM Studio and vLLM: one listener, one port, every engine's models in a single `/v1/models`, each request routed to the local engine that owns the requested model and each workload tagged with that engine.
- Manually added bare hosts are probed for SGLang on `30000`, told apart from other OpenAI-compatible servers by `GET /get_model_info`, a route only SGLang serves. The scheduler ranks SGLang nodes with the same node-wide load signal as the other engines.
- A tensor-parallel SGLang that spans several machines (`--nnodes N`) is adopted on its head rank only; worker ranks open no port and correctly show no engine. The docs say so, next to the equivalent vLLM note.

One SGLang-specific finding worth knowing: on current SGLang builds `GET /health` runs a real forward pass and takes about a second (the same as `/health_generate`), while `/get_model_info` answers in well under a millisecond. PAIR probes an engine's ready route often (status polls, model-list requests from every peer, the loaded-model watcher, the health loop), so every SGLang probe here uses `/get_model_info`. With `/health` a node with several peers serialized behind the engine's operation mutex and its model endpoint stopped answering; the last commit explains the measurement.

Related: #24 asks for adoption of OpenAI-compatible backends and names SGLang as one; this covers the SGLang case the way #9 covers vLLM, through the one shared OpenAI-compatible proxy rather than a new binary.

## Scope

Included: engine manifest (`services/nvpair-engine-manager/manifests/sglang.json`, Linux amd64/arm64, `uv` venv install of `sglang[all]`), discovery key `sg`, the OpenAI-compatible proxy's engine table (resolution order LM Studio, vLLM, SGLang), broker advertise loop and manual-node bridge, the bare-host probe leg with its `/get_model_info` disambiguator (and the matching guard so the vLLM leg never claims an SGLang), scheduler engine list, TUI label, desktop engine type, capabilities, icon and per-engine attribution, readable display of path-like model ids, docs, and tests for each of those.

Excluded: a managed port facade for SGLang (it keeps `30000` and is never moved), model download or deletion, any Windows or macOS install path, and authentication headers for an SGLang behind a token (#26).

**Stacked on #9 and #10.** SGLang needs #9's OpenAI-proxy generalization and served-model machinery, and #10's per-service port overrides for manually added nodes, so this branch is based on the merge of those two. Until they land, the diff here shows all three; the SGLang-only change is `https://github.com/cguldogan/Personal-AI-Router/compare/d676885...pr/sglang-engine` (8 commits, 61 files). No breaking IPC change: `node/add-manual` and `node/remove-manual` simply accept `sglang` as a third `engine` value.

## Validation

Environment: macOS 15 arm64, Go 1.27, Node 23; two DGX Spark (GB10, aarch64) nodes running a real two-rank SGLang group as the engine under test.

```
export PATH=/opt/homebrew/bin:$PATH
cd services && ./build.sh                       # 13 binaries, product 0.93.0
for m in shared lmstudio-proxy nvpair-ui-broker nvpair-manual-nodes nvpair-job-scheduler nvpair-tui nvpair-engine-manager nvpair-node-scanner; do
  (cd $m && gofmt -l . && go vet ./... && go test ./... -count=1)
done
cd services/tests && go test ./... -count=1     # cross-process interop, 331 s
cd desktop && npm run typecheck && npm run lint && npm run test:unit && npm run service-contracts:check && npm run dead-code:check
node scripts/spdx-headers.mjs                   # 896 checked, 0 missing
git merge-tree --write-tree upstream/main pr/sglang-engine   # clean
```

All pass: gofmt clean and vet clean in every module; 7 of the 8 Go modules fully green; the interop suite passes with 0 skips; desktop typecheck clean, lint 0 errors (one pre-existing prettier warning in a file this change does not touch), 243 unit tests in 42 files pass; service contracts and dead-code checks pass; every commit is signed off. The one failure is `nvpair-engine-manager TestUninstallTerminatesRunningInstance`, which fails identically on the unmodified base tree on this host (it cannot resolve the image path of the process holding a port; the same failure is noted in #9) and is unrelated.

A note for reviewers reading the series in order: the docs commit says SGLang readiness uses `GET /health`, and the final commit changes that to `GET /get_model_info` after measuring the cost against a real instance. Each message is accurate for its own commit; the end state is `/get_model_info` everywhere.

End-to-end against a real SGLang (`lmsysorg/sglang`, `--tp 2 --nnodes 2`, head rank on `30000`):

- The head node adopts it and shows SGLang installed, running and healthy; the worker rank shows no engine, as intended.
- The node advertises `sg` at its proxy port, its `/v1/models` carries the model attributed to `sglang`, and a chat completion sent to a **peer's** PAIR proxy routes over the cluster to the head node and answers.
- The same chat sent to the head node's own proxy answers locally, and to a macOS desktop node's proxy routes to the head node; the Jobs panel shows the run tagged SGLang.
- Eight concurrent model-list requests on the head node complete in about 10 ms total after the `/get_model_info` change; with `/health` they took 1 s each and serialized.

Screenshots of the node card (SGLang adopted on the head node) and of the Jobs panel (a run tagged SGLang) follow in a comment.

## Risk

- Additive IPC/HTTP only: a new discovery key, a new accepted `engine` value, new `sglang_*` manual-node status fields and a `sglang` port override. Consumers that ignore unknown keys and fields are unaffected; the desktop is updated in the same change.
- The probe route matters for SGLang specifically: reverting to `/health` reintroduces the one-second-per-probe cost described above. The manifest, the broker probe and the user doc all say why.
- Model ids from SGLang can be filesystem paths (leading slash, several segments). Nothing in the Go services parses model ids, they are compared by equality only, and this is covered by tests using such an id; the desktop renders the last path segment for display and keeps the raw id on the wire.
- No packaging, migration or data changes. The bundled manifest is Linux-only, so nothing changes on Windows or macOS beyond a listed, not-installable engine.

## Checklist


Noah-Tervalon-Nvidia:
Hey, thanks for adding this! We don't have a direction right now on adding SGLang, but will keep you posted.

jlacroix82:
One hazard to fold in from running this on a 4-node SGLang head for a week: on SGLang, `GET /health` is not a cheap liveness ping. It runs a short real generation (a 256-token prefill plus a few decode steps on the build I have, `lmsysorg/sglang:dev-dsv41` serving DeepSeek-V4.1-Flash) and takes ~1.0 s. This PR's manifest uses `/health` for `ready` and `health`, and the advertiser reuses `checkVLLMHealth`, so PAIR ends up issuing ~3,500 generations an hour against an idle head. Two effects:

- `engine:status` holds the engine lock for the ~1 s probe, and the broker/loaded-watch/desktop polls add up to about one status call per second, so `/v1/models` on that engine-manager never completes (details and code pointers in the issue I just filed).
- On that build the head leaked ~0.4 GB/h of front-end memory under that probe load until earlyoom killed the scheduler 20 h after boot; the three worker ranks stayed flat.

`/get_model_info` (the endpoint this PR already uses for identity) is ~1 ms, runs nothing, and its `model_path` field is a positive identity check the OpenAI proxy's facade can never satisfy. I switched all three probe sites to it in my fork, with a test that fails if the advertiser ever hits `/health`: https://github.com/jlacroix82/Personal-AI-Router/commit/c5b9be7. Happy to send it against this PR's branch if you would rather take it here.

Noah-Tervalon-Nvidia:

Issue #25, opened 2026-09-05
[Bug]: Action HTTP response-header timeout is selected by a hardcoded engine name, so only Ollama gets the long client


jlacroix82:
### PAIR version or commit

0.91.7 (nvpair-engine-manager 0.17.4, linux/arm64). The code paths below are also present on `main`.

### Affected component

Go services — nvpair-engine-manager

### Environment

NVIDIA DGX Spark (GB10, aarch64), Ubuntu 24.04.4 LTS, driver 580.173.02, 10-node PAIR cluster. Reproduced against stock binaries with no local modifications.

### Steps to reproduce

1. Configure an engine other than Ollama.
2. Issue an `engine:action` whose HTTP call needs more than 30 seconds to return response headers. The realistic case is a `chat` action against a large model with a long prompt, where prefill alone can exceed 30s before the first header is written.
3. Observe the action fail with a transport timeout.

### Expected behavior

The per-action timeout should be a property of the action or the manifest rather than of the engine's name, so an engine author can declare that a given action may be slow.

### Actual behavior

`executor.go` defines two clients:

```go
engineResponseHeaderTimeout     = 30 * time.Second
ollamaLoadResponseHeaderTimeout = 10 * time.Minute
```

and `actions.go` chooses between them with a literal engine-name comparison:

```go
client := e.client
if engine == "ollama" && action == "run_model" && e.ollamaLoadClient != nil {
    client = e.ollamaLoadClient
}
```

Every other engine therefore gets 30 seconds for every action, and no manifest field can change it.

By code inspection this is not limited to third-party engines. The shipped `lmstudio.json` declares a `chat` action (`POST /v1/chat/completions`) that uses the 30s client, and the vLLM manifest proposed in #9 declares a `chat` action as well, so a first-class vLLM engine would inherit the same ceiling.

The failure also presents badly: `net/http: timeout awaiting response headers` reads as a broken or hung engine rather than a client-side limit.

### Sanitized logs or screenshots

```
action "run_model": Post "http://127.0.0.1:8800/v1/chat/completions":
net/http: timeout awaiting response headers      [30s]
```

[...]

PR #52, opened 2026-09-09
ollama-proxy: make response-header timeout configurable

pettersandvand:
## Summary
`proxyResponseTimeout` (`services/ollama-proxy/proxy.go`) is a hardcoded 120s constant used as the `ResponseHeaderTimeout` on the transport that forwards real client requests (e.g. `/v1/chat/completions`) to a chosen backend node. There's no flag or env var to change it.

In practice this means any backend response that legitimately takes longer than 120s to start — a cold model load on a large local model, a long prefill, or an engine (e.g. Ollama) that only flushes output once a tool-call has fully composed — fails with:

```
net/http: timeout awaiting response headers
```

even though the backend is healthy and would have answered given more time. This reproduces reliably with larger local models (e.g. 27B+) doing tool-calling generations.

## Change
- Added a `--response-timeout` duration flag to `services/ollama-proxy/main.go` (default `120s`, matching current behavior exactly — no change unless explicitly set).
- Threaded the value through `Proxy` via a new `SetResponseTimeout` method, applied to both the plain local forwarding transport and the per-peer cluster-mTLS transport (`newProxyTransport`), since both were previously governed by the same hardcoded constant.
- Left the fast model-list probe client (`modelListClient`, fixed at `proxyDialTimeout` = 10s) untouched — it's intentionally a quick health/listing probe, not the request-forwarding path.
- Updated the README flags table.
- Added tests: default timeout applied, custom override applied to the transport, non-positive values ignored.

## Testing
- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./...` — all tests pass except one pre-existing, unrelated failure (a loopback-alias test requiring a `127.0.0.2` bind, which fails identically on unmodified `main` in this sandbox — confirmed via `git stash`).

## Notes
Happy to adjust the flag name/shape (e.g. accept it in ms instead of a Go duration string, or expose it per-engine rather than globally) if maintainers have a preferred convention.


pettersandvand:
Related: #25 — same underlying pattern (a hardcoded response-header timeout with no config surface), but a different service/code path. #25 is about `nvpair-engine-manager`'s internal action dispatcher (`executor.go`'s engine-name-gated 30s/10m client selection, used for install/pull/local action calls). This PR is about `ollama-proxy`'s separate `proxyResponseTimeout` constant, which governs the client-facing forwarding path (e.g. `/v1/chat/completions`) and is what actually causes real user-facing request failures on longer-running generations. Both stem from the same design gap (fixed timeouts, no way to override), so flagging the connection in case a maintainer wants a unified approach across both services.

sherief-nv:
Thanks - timeout units in minutes as a granularity seem too coarse, can we make them seconds?

Moving MR against development branch.

Noah-Tervalon-Nvidia:

PR #63, opened 2026-09-11
Remove CORS allow policy

kjlubick:
## Description

PAIR's inference proxy endpoints are intended for native clients, but the shared wildcard CORS policy allowed browser origins to send requests and read proxy or engine responses.

This removes that and updates tests and documentation to match.

This bumps `ollama-proxy` and `lmstudio-proxy` because of the new behavior.

## Scope

Blocking browser-based queries on our proxies is only layer in a defense-in-depth strategy. See SECURITY.md for more

## Validation

I ran tests locally and spun up a local 1-node instance to do inference to make sure one CUJ works.

```
cd services/shared
go test ./...

cd ../ollama-proxy
go test ./...

cd ../lmstudio-proxy
go test ./...
```

## Risk

There might be other CUJs I'm not familiar with that will break. If they do, we'll want to be more careful than just a wildcard Allow.

## Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/NVIDIA/Personal-AI-Router/blob/main/CONTRIBUTING.md).
- [X] Every commit is signed off (`git commit -s`), certifying the [Developer Certificate of Origin](https://developercertificate.org/).
- [X] New or existing tests cover the change.
- [X] Relevant documentation is updated.
- [X] I checked the diff, changed filenames, and commit messages for credentials, private data, internal URLs, internal issue identifiers, and generated artifacts.
- [X] I recorded the validation commands and results above.
- [X] I bumped any affected component in `services/versions.json`, and described user-visible changes above so they reach the release notes.


kjlubick:
Internal tests pass: https://gitlab-master.nvidia.com/dhc/personal-ai-router/-/pipelines/67784955

kjlubick:
Made obsolete by internal change, coming soon to the public

PR #70, opened 2026-09-12
fix: send mDNS responses from source port 5353

### LINKED ISSUE #1: [Bug]: mDNS answers are sent from an ephemeral source port, so reflectors never relay node records across subnets
### PAIR version or commit

`13b68115fa2c9c1d94f1ead1358f8d5a527cfecf` ("Update to 0.1.1"), built from source with `make build-services`. `nvpair-ui-broker --version` reports `0.40.2`; the services bundle banner reports `product v0.91.7`.

### Affected component

Discovery or pairing

### Environment

```
OS:               Ubuntu 24.04 LTS (headless, multi-user.target) x2; Arch-based rolling x1; desktop app on a fourth machine
Architecture:     arm64 (2 nodes), x86_64 (1 node)
GPU and driver:   Jetson Orin integrated GPU (arm64 nodes); 2x discrete NVIDIA GPUs (x86_64 node)
Engine/version:   n/a - the failure happens during discovery, before any engine is involved
Model:            n/a
Cluster size:     4 nodes, split across two routed subnets on separate VLANs, all joined to one cluster
Build:            Go 1.27.1, `make build-services`; headless nodes run `nvpair-tui` inside tmux
Network:          router mDNS proxy/reflector enabled between the two VLANs; unicast fully routed between them
```

### Steps to reproduce

1. Put PAIR nodes on two different IP subnets that sit on separate VLANs, with a router-level mDNS proxy/reflector enabled between them (an RFC 6762 style reflector, i.e. one that relays traffic it identifies as mDNS on UDP 5353).
2. Confirm unicast between the subnets is open: `curl http://<peer>:14318/v1/node-info` returns HTTP 200 in both directions.
3. Start PAIR on a node in each subnet and wait. The Nodes tab never lists the peer from the other subnet. Nodes on the *same* subnet discover each other normally.
4. Sniff mDNS on a host in each subnet (bind UDP 5353 with `SO_REUSEADDR|SO_REUSEPORT`, join `224.0.0.251`, so the sniffer coexists with the running responder), filtering to packets sourced from the router's own address:
   - the reflector relays `_nvpair-node._tcp.local` **queries** in both directions (`qd=1 an=0 flags=0x0000`, first labels `['_nvpair-node','local']`);
   - every relayed **answer** packet (`flags=0x8400`, 3-26 answers) contains only unrelated `.local` records. No `_nvpair-node` record ever crosses.
5. Isolate why. From a socket bound to UDP **5353** that has joined `224.0.0.251`, send a PTR query for `_nvpair-node._tcp.local` and record the source port of every answer received. Every `_nvpair-node` answer arrives from an ephemeral high port, never from 5353.

### Expected behavior

PAIR sends its mDNS answers from UDP source port 5353, as RFC 6762 section 6 requires. A reflector then recognizes them as mDNS responses and relays them, and nodes on reflected subnets discover each other without manual configuration.

### Actual behavior

Every mDNS answer PAIR sends originates from an ephemeral source port. A reflector never classifies those datagrams as mDNS, so it relays the queries but silently drops the responses, and cross-subnet discovery never completes. Same-link discovery is unaffected because the querier receives the datagram directly and does not care about the source port - only a reflector does.

The source of the behavior is `services/shared/mdns/responder.go`, in `sendOnInterface`. Every transmit binds a fresh socket with an ephemeral port:

```go
conn, err := net.ListenUDP("udp4", &net.UDPAddr{IP: src, Port: 0})
```

The function's own doc comment explains the intent:

> `sendOnInterface` is the core of the Windows send workaround: it transmits buf from a fresh unicast-bound socket on the given interface (setting the multicast interface + TTL for group targets), never from the multicast-bound receive socket that Windows refuses to send from.

The Windows workaround is reasonable, but applying it on every platform makes PAIR's responses non-conformant on all of them.

Suggested fix: on non-Windows platforms, transmit from the socket already bound to 5353, or bind the per-interface send socket to port 5353 with `SO_REUSEADDR`/`SO_REUSEPORT`, and keep the ephemeral-port path as a Windows-only fallback.

### How we confirmed the router's mDNS reflection was actually on

This was ruled out first, since a misconfigured reflector produces the same symptom.

- The gateway's mDNS proxy was enabled with a VLAN scope that included **both** the VLAN holding the desktop/UI node and the VLAN holding the headless nodes.
- The service scope was tried three ways: a specific allowlist, then a custom entry added for PAIR's service type, then set to **All**. The measurements above were re-run after each change and were identical every time.
- Positive control that reflection was live: the sniffers on **both** sides captured mDNS packets sourced from the gateway's own address, including relayed `_nvpair-node._tcp.local` queries that had originated on the opposite subnet. Multicast reflection was demonstrably working in both directions - only the answers were missing.
- Unicast routing between the subnets is fully open (HTTP 200 on `/v1/node-info` both directions), so this is not a firewall or routing problem.
- `avahi-daemon` was inactive on the headless nodes, so no second responder was competing with PAIR's own.
- PAIR advertises exactly one service type, `_nvpair-node._tcp` (`ServiceType` in `services/shared/noderec/noderec.go`), and that is the name the reflector was observed relaying queries for. The name is not the problem.

```shell
# Packets sourced from the gateway, captured on a host in each subnet.
# Queries for the PAIR service type are relayed across the VLAN boundary:
<gateway>  qd=1 an=0 flags=0x0000 labels=['_nvpair-node', 'local']
<gateway>  qd=1 an=0 flags=0x0000 labels=['_nvpair-node', 'local']

# Answers relayed by the gateway never contain a PAIR record:
<gateway>  qd=0 an=9  flags=0x8400 labels=['local']
<gateway>  qd=0 an=26 flags=0x8400 labels=['local']

# Query for _nvpair-node._tcp.local sent from a socket bound to UDP/5353.
# Source port of each answer received (expected 5353 per RFC 6762 section 6):
<local-node>:32809   _nvpair-node._tcp.local
<local-node>:33112   _nvpair-node._tcp.local
<local-node>:35195   _nvpair-node._tcp.local
<peer-node>:32845    _nvpair-node._tcp.local
<peer-node>:33047    _nvpair-node._tcp.local
<peer-node>:37499    _nvpair-node._tcp.local
```

Workarounds in the meantime: add each cross-subnet peer as a manual node (Manual tab, `a`), or place every PAIR node on a single VLAN.

### Confirmations

woodsonl:
## Summary

Fixes #1. mDNS answers were sent from a fresh socket bound to an ephemeral port, so a router-level mDNS reflector relayed PAIR's `_nvpair-node._tcp.local` queries but dropped every answer — it only classifies a datagram as mDNS when it originates from UDP 5353. Nodes on the same link still discovered each other; nodes on reflected subnets never did.

RFC 6762 §6 requires responses to originate from the well-known mDNS port. On Unix the responder now transmits from the same socket it already receives on, which is bound to 5353. No second socket is opened on 5353, so a unicast query addressed to the host is never captured by a short-lived send socket. Windows keeps a fresh per-send socket, since it refuses to send from the group-bound receive socket.

## Changes

- `services/shared/mdns/responder.go`: `Run` installs its receive socket as the send socket (`setSendPC`); `sendOnInterface` sends from it with a per-datagram `ipv4.ControlMessage{IfIndex}`, falling back to `sendOnFreshConn` on Windows and for sends before `Run` installs it. Sets the multicast TTL once on each send socket with `SetMulticastTTL` — `ControlMessage.TTL` is receive-only in `x/net/ipv4` — via a named `mdnsTTL` constant.
- `services/shared/mdns/socketreuse_unix.go` / `socketreuse_windows.go`: add `sendFromRecvSocket` (true on Unix, false on Windows). The fallback send socket always binds an ephemeral port, so it can never capture unicast addressed to the 5353 receive socket.
- `services/shared/mdns/responder_test.go`: assert the fallback uses an ephemeral port, coexists with the group socket, errors when the interface has no addresses, and that a send uses the installed shared socket.
- `services/versions.json`: bump `nvpair-node-scanner` 0.20.3 → 0.20.4 (the only consumer of the shared responder).

## Validation

- `go test -race -count=1 ./mdns/` in `services/shared` — passes, including the send-socket tests (ran, not skipped).
- `go test -count=1 ./...` and `go build ./...` in `services/nvpair-node-scanner` — pass.
- Cross-compiles clean for `windows/amd64`, `darwin/arm64`, and `linux/arm64`.
- Live bind probe: a unicast datagram to `src:5353` goes to the most-specific socket, which is why the responder sends from its receive socket instead of a fresh one; the fallback is kept off 5353 so it cannot steal.

## Notes

No JSON-RPC, configuration, or user-facing surface changes. Cross-subnet discovery needs no manual peers once the reflector is in the path.

### DIFF
--- a/services/shared/mdns/responder.go
+++ b/services/shared/mdns/responder.go
@@ -16,9 +16,11 @@
 // invisible to LAN peers.
 //
 // We keep zeroconf's receive trick (join the group on each multicast interface,
-// which works fine on Windows) but send every reply/announcement from a
-// per-interface unicast-bound socket with SetMulticastInterface set explicitly.
-// That path is well-supported on Windows.
+// which works fine on Windows). On Unix, replies and announcements go out that
+// same receive socket, so they originate from 5353 as RFC 6762 §6 requires. On
+// Windows, which refuses to send from a group-bound socket, they instead go out
+// a per-interface unicast-bound socket with SetMulticastInterface set
+// explicitly; that path is well-supported there.
 //
 // This is the single implementation consolidated (the mDNS dedup) from the five
 // near-identical copies that lived in nvpair-advertiser,
@@ -97,6 +99,14 @@ type Responder struct {
 	addrMu sync.RWMutex
 	// ifaceAddrs maps interface index to its IPv4 unicast addresses.
 	ifaceAddrs map[int][]net.IP
+
+	// sendMu guards sendPC. Run sets it on platforms that transmit from the
+	// receive socket (see sendFromRecvSocket); while non-nil, sends go out
+	// that socket so they share its source port and never briefly bind a
+	// second socket to 5353, which would capture unicast queries meant for
+	// the receive socket. Nil on Windows and before Run binds.
+	sendMu sync.Mutex
+	sendPC *ipv4.PacketConn
 }
 
 // NewResponder builds a Responder for the given service instance. domain
@@ -155,6 +165,22 @@ func (r *Responder) ifaces() map[int][]net.IP {
 	r.addrMu.RLock()
 	defer r.addrMu.RUnlock()
 	return r.ifaceAddrs
+}
+
+// setSendPC installs (or clears) the socket sends use instead of a fresh
+// per-send socket. Called by Run on platforms where sendFromRecvSocket is set.
+func (r *Responder) setSendPC(pc *ipv4.PacketConn) {
+	r.sendMu.Lock()
+	r.sendPC = pc
+	r.sendMu.Unlock()
+}
+
+// sharedSendPC returns the shared send socket, or nil when sends should open a
+// fresh per-send socket (Windows, or before Run binds).
+func (r *Responder) sharedSendPC() *ipv4.PacketConn {
+	r.sendMu.Lock()
+	defer r.sendMu.Unlock()
+	return r.sendPC
 }
 
 // UpdateTXT swaps the advertised TXT records and re-announces immediately. It
@@ -257,6 +283,14 @@ func (r *Responder) Run(ctx context.Context) error {
 	// but functionally correct. So we treat any error from this as advisory.
 	if err := pc.SetControlMessage(ipv4.FlagInterface, true); err != nil {
 		slog.Debug("mdns: control message not available, will reply on all interfaces", "reason", err)
+	}
+
+	// Transmit from this same socket on platforms where that is allowed, so
+	// responses leave from 5353 (RFC 6762 §6) without a second socket on 5353
+	// that would capture unicast traffic meant for this one.
+	if sendFromRecvSocket {
+		r.setSendPC(pc)
+		defer r.setSendPC(nil)
 	}
 
 	var joined int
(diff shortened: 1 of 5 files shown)


Noah-Tervalon-Nvidia:
See note above, going to close this as a dupe, but we will make sure to give you a mention in our next release notes! Thank you so much for the ideas here, we added some minor stuff and some scaffolding around these ideas, and while in the future we would have worked with you on these fixes we wanted to get them in earlier so we were working off our internal set of changes and just weren't slide them in on that side before we had the large push out yesterday.

Issue #19, opened 2026-09-05
[Bug] nvpair-node-scanner causes macOS kernel panics (userspace_watchdog_timeout) due to aggressive net.Interfaces() polling


CoDev-BitByBit-Commit:
## System Information
- **OS:** macOS (Apple Silicon / arm64)
- **Component:** `nvpair-node-scanner` / `shared/discovery`
- **Symptom:** Entire system freezes and reboots with a `userspace_watchdog_timeout` kernel panic.

## Description
When running the PAIR macOS application, the system will randomly freeze and trigger a forced watchdog reboot. Analyzing the Apple Diagnostic Reports (`/Library/Logs/DiagnosticReports/configd_*.spin`) reveals that Apple's core network configuration daemon (`configd`) gets starved while waiting for a kernel read-write lock (`krwlock`). 

The macOS crash logs explicitly show that this kernel lock is being held by `nvpair-node-scanner`:

```text
*1 ??? (kernel.release.t6050 + 773428) [0xfffffe000b4a8d34] (blocked by krwlock for reading owned by nvpair-node-scanner [1309] [unique pid 791059] thread 0x6fa35b) 2
```

## Root Cause Analysis
The root cause stems from the Windows mDNS workaround in `services/shared/discovery/discovery.go`. 

The `Browse` loop executes every 5 seconds. Within this loop, `sendMulticastQuery()` is called to bypass a Windows-specific bug where `grandcat/zeroconf` drops UDP packets due to multicast bindings. However, `sendMulticastQuery()` executes unconditionally on **all** operating systems.

```go
func sendMulticastQuery(service, domain string) map[string]bool {
    // ...
    ifaces, err := net.Interfaces()
    // ...
    // Iterates through all interfaces, binding and closing UDP sockets
```

On macOS, calling `net.Interfaces()` and rapidly binding/closing raw UDP sockets on every valid interface every 5 seconds inside a background daemon causes severe kernel routing table lock contention (`krwlock`). Over time, this starves `configd`, triggering the `userspace_watchdog_timeout` panic.

As the source comment itself notes, standard `zeroconf` receive/transmit works perfectly on macOS/Linux. This custom per-interface UDP barrage is only necessary on Windows.

## Proposed Fix
Guarding the `sendMulticastQuery()` function so that it only executes on Windows completely eradicates the kernel panics on macOS, while preserving standard `zeroconf` discovery.

**Patch for `services/shared/discovery/discovery.go`:**

```go
[...]

PR #72, opened 2026-09-13
Resolve multiple mDNS issues and macOS network stack degredation

### LINKED ISSUE #59: [Bug]: WiFi saturation on wireless host node / WLAN driver deadlock
### PAIR version or commit

0.1.1 (0.1.1.463)

### Affected component

Desktop application

### Environment

OS: macOS 26.6.2 (25G83) on affected node (wireless host); peers run macOS (current)
Architecture: arm64 (Apple Silicon, Mac16,8) on all nodes
GPU and driver: Affected node — Apple Silicon integrated GPU (no NVIDIA GPU). WLAN driver: com.apple.DriverKit-AppleBCMWLAN (system DriverKit extension). Peers — Apple Silicon integrated GPUs.
Engine and version: PAIR 0.1.1 (0.1.1.463, com.nvidia.nvpair); engines: LM Studio 0.4.24 (headless llmster daemon) + Ollama on all nodes
Model: N/A on affected node (no model loaded); issue occurs with cluster idle — no inference running
Cluster size: 5 nodes (4 on wired Ethernet, 1 on WiFi — the affected host is the only wireless node)

### Steps to reproduce

1. Set up a 5-node PAIR cluster: 4 nodes on Ethernet, 1 node on WiFi. Pair and trust all nodes.
2. Leave the cluster idle (no model loads, no jobs).
3. On the WiFi node, launch PAIR and leave it running for 8–10 minutes.
4. From the WiFi node, ping the local gateway continuously (1/s) while PAIR runs.
5. Observe latency for ~10 minutes, then quit PAIR and keep pinging.

### Expected behavior

The wireless host's network stays healthy while PAIR runs: gateway latency stays in the single-digit
milliseconds with 0% loss, on any SSID. Cluster background traffic (discovery, sync, health polls)
should not meaningfully degrade the host's connectivity.

### Actual behavior

After ~8 minutes of PAIR running on the wireless host, gateway latency explodes in continuous bursts
regardless of which SSID is selected:

  Healthy (PAIR just launched):   3–12 ms, 0% loss
  Degraded (after ~8 min):        bursts up to 1,441 ms (also 1,169 / 1,118 / 1,085 / 1,048 ms...),
                                  internet pings degrade in mirror -> bottleneck is the local link
  Quit PAIR:                      instant recovery to 3–7 ms, 0% loss (reproduced twice)

In a worst case the same day, the WLAN driver itself wedged and was killed by macOS:
com.apple.DriverKit-AppleBCMWLAN crashed (SIGABRT, thread "WLANDextWatchdogDQ") with stack
IO80211DeadlockPolice::handleDextWatchdogTimer() -> DriverKit panic -> abort() —
a hard WiFi outage until the dext restarted. This was the only WLAN crash in 72h and never
reproduced outside PAIR execution.

```shell
1) Ping timeline (gateway, 1/s, local times)
   11:48–11:55  PAIR running, healthy: 3–12 ms
   11:56–11:58  PAIR running, degraded: sustained bursts 300–1,441 ms; 0 timeouts
   11:58        PAIR quit -> recovery within one ping sample

2) PAIR cluster traffic during the degraded window (from nvpair.jsonl, 5 minutes):
   ~750 "cluster:trust-changed" events
   ~165 remote engine-status polls
   ~38 mDNS discovery cycles
   i.e. cluster writes fan-out to 4 peers every ~500 ms + multicast discovery, with
   zero PAIR-side errors (peers kept responding — the radio link itself saturated).

3) WLAN driver crash report (worst case, same day):
   Process:      com.apple.DriverKit-AppleBCMWLAN (system DriverKit extension)
   Termination:  SIGABRT (abort() called), thread "WLANDextWatchdogDQ"
   Crash stack (top frames):
     libsystem_kernel.dylib __pthread_kill
     libsystem_pthread.dylib pthread_kill
     libsystem_c.dylib abort
     DriverKit panic
     IO80211DriverKit IO80211DeadlockPolice::handleDextWatchdogTimer() (.cold.1)
     IO80211DriverKit IO80211DeadlockPolice::handleDextWatchdogTimer()
     IO80211DriverKit IO80211DQTimer::TimeOutOccurred_Impl(OSAction*, unsigned long long)
     DriverKit IOTimerDispatchSource::TimerOccurred_Invoke(...)
   Last DISASSOC logged ~35 s after crash capture; cascade of re-associations followed.

4) Ruled out:
   - PAIR installs no kernel/network system extension (pure userspace: app + launchd helper
     listening on no port + userspace proxies). Helper CPU ~0% at rest.
   - No interface errors on the wireless interface all day (0 Ierrs/Oerrs).
   - Issue does not occur on the 4 wired nodes (identical PAIR versions).
```

### Confirmations

stevbev:
## Description

Closes #59 #19 #1

Resolves macOS network stack degradation caused by two primary issues:
- opening new sockets for UDP:5353 for every broadcast and never closing the socket, leading to excessive socket binds that severely degrade the network stack over time
- a race condition caused by other macOS apps that bind mDNS port UDP:5353 shared, but Nvidia PAIR repeatedly binds UDP:5353 exclusive

## Scope

The change scope focuses on the mDNS implementation:
- replacing grandcat/zeroconf with a local mDNS solution that does not continuously open new sockets for broadcasts.
- on macOS binds UDP:5353 with SO_REUSEPORT to enable multiple apps to use the port, such as macOS system mDNSResponder and user programs like Chrome.
- limits macOS network interfaces used to broadcast to ones with a valid IPv4 address, and explicitly excludes the Airdrop (awdl) and low latency (llw) interfaces because those are invalid targets for PAIR.
- Feature gates the Windows broadcast workaround to only run on Windows devices, so that the double broadcast on unix systems is resolved.

## Validation

The code update was tested on a 4-node Nvidia PAIR cluster (3 macOS, 1 Windows 11), with the affected macOS node experiencing severe network issues on PAIR v0.1.1 having no issues on this PR after 7 days of operation.

## Risk
- The features of grandcat/zeroconf used in PAIR v0.1.1 were minimal, so there aren't any expected negative impacts from replacing grandcat/zeroconf with a local minimal mDNS solution.
- There might be some edge use cases where macOS users are trying to use the Airdrop or low-latency-direct-connection interfaces with PAIR. Since these WiFi access modes require exclusive access to the network that prevents normal traffic to the network, the risk should be minimal. Removing the network interface filtering will eliminate this risk with a minor increase in network interface scan processing.

## Checklist


mkalkere (commented):
Good fix for the socket churn -- pooled per-interface send socket plus one long-lived receiver is the right shape.

Blocking concern: the new receiver binds 224.0.0.251:5353 without SO_REUSEADDR/SO_REUSEPORT, while the responder in the same process and the system mDNSResponder already hold that port with reuse options. On macOS the receiver bind will likely fail with EADDRINUSE and every scan returns empty -- discovery goes fully dark on the platform this is fixing. Same on Linux where node-scanner and both proxies each run a browser: first bind wins, the rest go dark. `NewReceiver` needs the same reuse-address control hook as the responder (with SO_REUSEPORT on darwin) before merge.

Non-blocking: `net.Interfaces()` polling cadence is unchanged, the receive path is IPv4-only, and per-datagram assembly drops cross-packet aggregation. Minor: despite the branch name this is a hand-rolled client rather than native mDNS, but that's cosmetic.

stevbev:
> Good fix for the socket churn -- pooled per-interface send socket plus one long-lived receiver is the right shape.
> 
> Blocking concern: the new receiver binds 224.0.0.251:5353 without SO_REUSEADDR/SO_REUSEPORT, while the responder in the same process and the system mDNSResponder already hold that port with reuse options. On macOS the receiver bind will likely fail with EADDRINUSE and every scan returns empty -- discovery goes fully dark on the platform this is fixing. Same on Linux where node-scanner and both proxies each run a browser: first bind wins, the rest go dark. `NewReceiver` needs the same reuse-address control hook as the responder (with SO_REUSEPORT on darwin) before merge.
> 
> Non-blocking: `net.Interfaces()` polling cadence is unchanged, the receive path is IPv4-only, and per-datagram assembly drops cross-packet aggregation. Minor: despite the branch name this is a hand-rolled client rather than native mDNS, but that's cosmetic.

Thank you for the feedback on the PR. I am a little puzzled though because I double checked and the new receiver binds to port 5353 using SO_REUSEADDR and SO_REUSEPORT on macOS, here in the new [socketreuse_darwin.go](https://github.com/stevbev/Personal-AI-Router/blob/076aa496ba5f43a38462c6459c59345438913372/services/shared/mdns/socketreuse_darwin.go#L22). Can you help identify which line of code you think is binding the port without SO_REUSEPORT?

Here is a screenshot showing that port 5353 is bound shared with Chrome and PAIR using it.
Screenshot 2026-09-14 at 9 16 49 PM

stevbev:
I reviewed the branch again and saw that I completely missed the test files still using zeroconf, which is a glaring inconsistency between the service code and the test cases. The latest commit on the PR finishes removing all the grandcat/zeroconf invocations in the code base so it is a clean implementation and proper testing of the new mDNS module.

Noah-Tervalon-Nvidia:

stevbev:
This PR can no longer be merged because there is a competing solution on https://github.com/NVIDIA/Personal-AI-Router/commit/bda6d875f90d9eb07a7bb7aef8de781c3d80bde4.

PR #87, opened 2026-09-15
ci: run checks and builds on GitHub Actions, with release-intent automation

Noah-Tervalon-Nvidia:
## Description

Moves checks and unsigned installer builds to GitHub Actions on GitHub-hosted
runners, and ports the release-intent automation that keeps version bumps and
changelog entries out of merge-conflict territory.

Self-hosted runners cannot serve a public repository: on a fork pull request
the workflow file comes from the contributor's merge ref, so `runs-on` is
attacker-chosen. Everything here runs on hosted runners and holds no secret,
so there is no credential for pull request code to reach. Signed releases stay
on internal infrastructure and are built from a tag.

Also renames the `versions.json` field `product` to `services`. It only ever
versioned the services suite — it stamps the standalone installer and Go
`main.Version`, while the version users install lives in
`desktop/package.json`, which was explicitly out of scope for it. The two had
drifted to `0.91.7` and `0.1.1`.

Resolves https://github.com/NVIDIA/Personal-AI-Router/issues/22

### Changelog title
n/a

### Changelog body
n/a

### Bumps
- services: none
- lmstudio-proxy: none
- nvpair-cluster-manager: none
- nvpair-engine-manager: none
- nvpair-errors: none
- nvpair-job-scheduler: none
- nvpair-manual-nodes: none
- nvpair-node-info: none
- nvpair-node-scanner: none
- nvpair-node-settings: none
- nvpair-tui: none
- nvpair-ui-broker: none
- nvpair-workload-manager: none
- ollama-proxy: none

## Scope

Included: workflows, the release-intent scripts, the `product` to `services`
rename across all nine consumers, a changelog seeded from the published 0.1.0
and 0.1.1 release notes.

Excluded: signing, publishing, and release. Those stay on internal
infrastructure and are a separate piece of work.

## Validation

Locally, on this branch:

- `python3 scripts/release-intent/test_lib.py` — 28 tests
- `npm --prefix desktop run typecheck` — clean
- `npm --prefix desktop run lint` — 0 errors
- `npm --prefix desktop run dead-code:check` — no dead code
- `npm --prefix desktop run test:unit` — 208 tests
- `actionlint` on all four workflows — clean
- `node scripts/spdx-headers.mjs` — 0 missing
- `apply_pr.py --dry-run` produces `0.1.1 -> 0.1.2` and a correctly numbered
  changelog section

No workflow has executed before this pull request. The six-way build matrix is
unproven, and two legs are worth watching: Windows builds natively rather than
cross-compiled under wine, and `macos-latest` is the smallest box in the matrix
at 3 cores and 7 GB.

## Risk

All bumps are `none`: no Go source changed, so no compiled output changes and
this is not a release.

`services/versions.json` and `CHANGELOG.md` are modified, which the new
bot-owned path check rejects by design. The `allow-owned-files` marker at the
top of this description is the documented override for a schema change the bot
cannot make itself.

`Release intent apply` does not run on a pull request. It first runs on merge
to `develop`, and needs the GitHub App configured before then or it will fail.

## Checklist

- [x] I have read the Contributing Guidelines.
- [x] Every commit is signed off (`git commit -s`).
- [x] I checked the diff, changed filenames, and commit messages for
      credentials, private data, internal URLs, internal issue identifiers, and
      generated artifacts.
- [x] I declared version bumps in the release-intent block above.


.github/workflows/release-intent-apply.yml
```diff
@@ -0,0 +1,53 @@
...
+on:
+    push:
+        branches: [develop]
+
+permissions:
+    contents: read
+
+# Serializes applies so two merges cannot race to move the ref. The script also
+# retries a non-fast-forward, so this is belt and braces.
+concurrency:
```
ckelseynv, line 28: I want to just double check on a potential issue: I don't think this `concurrency` block does what the comment says, and I think it
can quietly eat a release.

GitHub Actions holds only **one** pending run per concurrency group. When a third
run shows up, it cancels whichever one was already waiting. `cancel-in-progress:
false` protects the run that's actively executing; it does nothing for the one
sitting in the queue behind it. So this group doesn't serialize applies — it
drops them.

What makes this more than theoretical is that the job triggers itself. The bot
pushes to `develop` with an App installation token, and unlike `GITHUB_TOKEN`
those pushes *do* start workflow runs. So: merge A starts a run, merge B queues
behind it, then A's own `[release-intent]` commit lands and queues a third run —
which evicts B. B's version bump and changelog entry are just gone. No failure,
no retry, just a cancelled run in the Actions tab that nobody is watching.

The good news is `apply_pr.py` doesn't need the serialization, and the comment
here already half says so. It reads the branch head, builds the tree against it,
and PATCHes the ref with `force: false`, so a concurrent apply comes back 409/422
and the loop re-reads and retries. That retry *is* the concurrency control. The
group on top of it isn't belt and braces, it's an extra failure mode.

Suggest dropping the block and letting each push get its own run:

    # Concurrent applies are handled by the script, not here: it commits with
    # `force: false` and re-reads the ref when the branch moved underneath it.
    # Do not add a `concurrency:` group — GitHub keeps only one pending run and
    # cancels any earlier one, so a burst of merges silently loses bumps.

Optional while you're in here: the bot's own push boots a full runner just to hit
the `[release-intent]` check in `main()` and exit. A job-level guard skips it
without starting anything:

    jobs:
        apply:
            if: ${{ !startsWith(github.event.head_commit.message, '[release-intent]') }}

.github/workflows/release-intent-check.yml
```diff
@@ -0,0 +1,49 @@
...
+# included, on every typo fix in a description. This check is cheap, so it gets
+# its own trigger.
+#
+# Needs no secret: the body arrives in the webhook payload, so this runs on
+# pull requests from forks like every other check.
+name: Release intent check
+
+on:
+    pull_request:
+        branches: [main, develop]
```
ckelseynv, line 20: Just want to double check, since `release-intent-apply.yml` is only triggered on develop, we should never expect to run on main nor push anything that has not gone through develop? I ask in the (hopefully rare) cases  we need to hotfix.

For agent:

The branch lists in these two workflows don't line up, and the gap silently
swallows release intents.
This check gates pull requests targeting `main` **and** `develop`, but
`release-intent-apply.yml` only triggers on pushes to `develop`. So a pull
request that targets `main` is required to fill in the intent block, is blocked
by `check_forbidden_paths` from touching `services/versions.json` by hand — and
then has its declared bumps thrown away on merge. No failing job, no warning,
nothing in the log. The contributor did everything the template asked and their
version bump just doesn't exist.
That isn't a hypothetical path either. `CONTRIBUTING.md:267` tells people to
"Create a focused branch from `main` in your fork," and the file never mentions
`develop` at all. So the documented workflow leads straight into the hole.
The second half of this is the release cut. A `develop` -> `main` pull request
carries the bot's own edits to `services/versions.json` and `CHANGELOG.md`, which
are exactly the two paths `check_forbidden_paths` rejects. Every single cut will
fail this check unless the description includes
`<!-- pair-release-intent-allow-owned-files -->`, and that marker isn't mentioned
in any contributor-facing doc — it only exists in the parser.
I don't think the answer is to make apply run on `main`. The reasoning in
`release-intent-apply.yml` for why it can't is right: a cut is one push covering
many pull requests and the `Applies-PR` trailer assumes one merge, one apply. The
conclusion to draw from that is to stop *accepting* release intent on `main`
rather than accepting it and dropping it.
Suggest dropping `main` from the branch list here, which fixes both problems in
one edit — no silently discarded bumps, and the release cut stops tripping the
bot-owned path check:
    on:
        pull_request:
            branches: [develop]
            types: [opened, synchronize, reopened, edited]
Two things to pair with that, so contributors don't end up on `main` in the first
place:
- Set the repository's default pull request base branch to `develop`, otherwise
  every fork PR still opens against `main` by default and skips the gate
  entirely.
- Update `CONTRIBUTING.md` to name `develop` as the integration branch. Right now
  step 2 of the pull request checklist actively points the wrong way, and `ci.yml`
  is the only place that says "Pull requests target `develop`."

services/VERSIONING.md
```diff
@@ -41,24 +52,29 @@ We follow [SemVer](https://semver.org/) (`MAJOR.MINOR.PATCH`).
...
 
 ## Declaring a version change
 
-Update `versions.json` in the same pull request that changes compiled output,
-using the tables above to pick the severity. Say in the pull-request description
-which components you bumped and why, and describe any user-facing change in plain
-terms so it can be carried into the release notes. A reviewer should be able to
-see the version decision without inferring it from the diff.
+Declare bumps in the `pair-release-intent:v1` block in your pull request
+description, using the tables above to pick the severity. **Do not edit
```
ckelseynv, line 70: This line introduces "Do not edit services/versions.json or CHANGELOG.md by hand", while AGENTS.md:127, CONTRIBUTING.md:249, and desktop/docs/services-backend.md:314 all still say to bump components in services/versions.json in the same pull request. services/readme.md and the pull request template were updated in this branch; these three were not. Note the internal validate:release-intent job is marked "GitLab-only; stripped on public cut", so this rule is genuinely new to public contributors and the docs are the only thing telling them about it.

.github/workflows/build.yml
```diff
@@ -0,0 +1,105 @@
...
+name: Build
+
+on:
+    workflow_dispatch:
+    push:
+        branches: [main, develop]
+    pull_request:
+        branches: [main, develop]
+        paths:
+            - '.github/workflows/**'
```
ckelseynv, line 20: The installer path filter covers .github/workflows/** but not .github/actions/**. A pull request that only bumps Node or Go in the composite setup action runs ci.yml on the new toolchain and skips the six-way installer matrix — precisely the cross-workflow drift the composite action exists to prevent. Add .github/actions/**

desktop/scripts/build-modular-binaries.ts
ckelseynv, line 313: I'd add `if (manifest.services !== versions.services) return false` as an existing cli-bin/manifest.json still counts as current, so the Services version silently disappears from the UI in dev.

mkalkere (approved):
Ran the release-intent scripts locally off this branch. test_lib.py: 27/28 (the 28th needs a repo checkout since it reads the real versions.json). validate_pr.py passes on this PR's body. apply_pr.py --dry-run with a release intent gives 0.1.1 -> 0.1.2, services 0.91.7 -> 0.91.8, package.json touched on one line, changelog entry lands in the right spot. Matches the validation section.

A few things:

1. apply_pr.py only applies the intent for the head commit's PR. If one push to develop ever carries two merges (direct push of a range, merge queue batch), the second intent is silently skipped and nothing fails. The description covers this for main cuts but develop has the same hole. Either iterate the push's commits or document that develop doesn't take batched pushes.

2. _parse_section ends a section at the first line starting with "### ". A changelog body containing a literal "### " line gets cut silently. Minor, a doc note would do.

3. The app plus RELEASE_INTENT_APP_ID / RELEASE_INTENT_APP_PRIVATE_KEY have to exist before the first push to develop, or the apply job fails. It fails loud (exit 2) and the README covers backfill, so not a defect, but I'd make it a pre-merge checkbox in the description.

4. The whole chain assumes branch protection requires the release-intent check. Admin-merging around it with a bad body fails loud at apply (exit 1) and versions just don't bump. Fine, but worth stating as an assumption.

Checked: macos-15-intel is a valid label (good until the macOS 15 image retires in fall 2027), so the mac-x64 leg will run. The product -> services rename covers all nine consumers including both sides of the IPC type, and dropping the installer key is deliberate and tested. Token scoping looks right: one-hour app token, contents:write, no Workflows permission, apply only on push to develop.

.github/workflows/release-intent-apply.yml
Noah-Tervalon-Nvidia: Confirmed and fixed in b0c3815f. You're right on both halves: GitHub keeps one pending run per group and cancels the earlier one, so cancel-in-progress: false protects the running job and does nothing for the queued one — and because an app installation token push does start workflow runs, the bot's own commit is the third run that evicts a real merge.

Dropped the block and took your comment close to verbatim so the next person doesn't re-add it. Also took the optional guard:

`if: ${{ !startsWith(github.event.head_commit.message, '[release-intent]') }}`

That skips the bot's own push without booting a runner, which removes the third run that made the eviction likely in the first place.

Also recorded "do not add a concurrency group here" as a known limit in scripts/release-intent/README.md, so the reasoning survives outside this thread.

.github/workflows/release-intent-check.yml
Noah-Tervalon-Nvidia: Answering the hotfix question first: no, nothing should reach main except a develop cut.

Fixed in b0c3815f — the check is now branches: [develop]. Your reasoning for why apply can't run on main is the same reasoning I'd used for its trigger, so I agree the conclusion is to stop accepting intent on main rather than accept it and drop it.

Both pairings are done in e0356582: CONTRIBUTING.md step 2 now says to branch from develop and open against develop, and explains that main receives develop as a release cut. It also documents the allow-owned-files marker, which as you noted existed only in the parser.

Checking with the team about swapping default branch. Only downside is see is people landing on dev when reaching it for the first time.

services/VERSIONING.md
Noah-Tervalon-Nvidia: Confirmed all three, fixed in e0356582. AGENTS.md, CONTRIBUTING.md, and desktop/docs/services-backend.md now say to declare bumps in the intent block and not to hand-edit the two bot-owned files.

Your point about the internal job being stripped on the public cut is what made this more than a tidy-up: for public contributors the rule is genuinely new, these files were the only thing that would have told them, and they were actively saying the opposite.

CONTRIBUTING.md now also documents the allow-owned-files override, since a contributor adding or removing a component has no other way past the check.

.github/workflows/build.yml
Noah-Tervalon-Nvidia: Fixed in b0c3815f, added .github/actions/**.

That's the exact hole: the composite action exists to stop ci.yml and build.yml drifting apart on toolchain versions, and the filter let a change to that action run one and skip the other.

desktop/scripts/build-modular-binaries.ts
Noah-Tervalon-Nvidia: Fixed in 2d87948a, and I reproduced the behaviour you predicted rather than just adding the line.

Built a manifest, then bumped only services in versions.json and re-ran without --force:

before: cli-bin is current … skipping build, and services: 0.91.7 stayed in the manifest
after: rebuilds and rewrites it
That stale value is what readCliBinManifest feeds the versions card, so it was exactly the silent-disappearance case you described.

Noah-Tervalon-Nvidia:
@mkalkere 

Thanks for running it end to end; matching numbers on the dry run is reassuring.

All four are real. Three are now documented as known limits in scripts/release-intent/README.md (e0356582) rather than fixed.

Batched pushes — agreed, but since I'll be the one merging everything (along with the team) for now I think not worth the fixes to get there since we will be doing squash commits.

'###' truncation — documented in the README and as a comment in the PR template, right next to the fence where an author would hit it.

Credential before first push — added to the README's credential section, and I've already verified it, but merging is the definitive check.

Branch protection assumption — stated in the README, including the admin-merge behaviour: fails loud at exit 1, versions simply don't bump.

mkalkere:
Verified the follow-ups — b0c3815f, e0356582, and 2d87948a all contain exactly what's described above. The diffs read clean.

Documenting the four findings as known limits rather than fixing them is a reasonable call, and the batched-push deferral makes sense while merges stay manual. Only thing I'd keep an eye on: the batched-push case is the one I'd least want to rediscover the hard way, since the skip is silent. It's documented now, so that's a conscious trade rather than a trap — just flagging it in case merge-queue batching ever comes up.

Thanks for the thorough response.

ckelseynv (approved):

PR #43, opened 2026-09-07
feat(node-info): report Linux AMD GPU utilization

dbourdea:
## Description

Linux AMD adapters can already execute GPU-backed inference through PAIR's existing Ollama integration, but node-info does not provide their live GPU utilization. This change reads the amdgpu `gpu_busy_percent` sysfs counter and matches it to detected AMD adapters by PCI address, so utilization is available to PAIR's existing telemetry and scheduling path.

An isolated two-node experiment on Ryzen AI Max+ 395 / Radeon 8060S systems verified Ollama 0.33.3 native ROCm gfx1151 execution, all 29 layers of Qwen3 0.6B offloaded, discovery, PIN pairing, model inventory and concurrent independent requests on distinct nodes. Inference worked with configuration only; this patch adds utilization reporting, not a new engine adapter.

Detailed findings, six reproducible figures, sanitized measurements, failure cases and setup notes: [public experiment repository](https://github.com/dbourdea/pair-amd-strix-halo-paper) and [white paper DOI](https://doi.org/10.5281/zenodo.22640202).

## Scope

- Recognize AMD PCI vendor 0x1002 and normalize PCI addresses independently of DRM card numbering.
- Accept genuine idle zero and valid utilization through 100%; leave missing or malformed counters unavailable.
- Preserve existing NVIDIA samples and append ghw AMD adapters when the NVIDIA detection path succeeds.
- Keep AMD memory fields absent because reserved VRAM, GTT, HSA-visible capacity and host MemAvailable have different, overlapping meanings.
- Add focused sysfs fixture tests and documentation; bump node-info from 0.13.3 to 0.14.0.
- No RPC contract, desktop, engine adapter, driver, process ownership or network security changes.

## Validation

On Ubuntu 26.04.1, Go 1.26.0, AMD gfx1151:

```sh
cd services
./build.sh
cd nvpair-node-info
GOMAXPROCS=4 go test -race ./...
```

All 13 service executables built. Ordinary `go test ./...` passed in shared, engine-manager, ui-broker (including relay/workloadstore), ollama-proxy, lmstudio-proxy, scheduler, scanner and cluster-manager components. Saved logs are in the linked evidence repository. Windows amd64 and macOS arm64 node-info cross-compilation with CGO disabled passed; these are build checks only.

[...]

PR #99, opened 2026-09-20
node-info: AMD GPUs on Linux from amdgpu sysfs (name, unified pool, busy %)

dmmdea:
## Intent
A Radeon host on Linux has no `nvidia-smi`, so node-info fell through to `ghw` and published the PCI database's codename and nothing else (`{"name":"Barcelo"}`): no capacity, no usage, no utilization, `telemetryValid` never true. The amdgpu driver already exposes everything the row needs in sysfs.

## What changed (`services/nvpair-node-info`)
- `gpu_amd_linux.go`: enumerates `/sys/class/drm/card<N>/device` with vendor `0x1002`; marketing name from a small APU/GPU device-id table (ghw's name as the fallback); capacity from `mem_info_vram_total` (+ `mem_info_gtt_total` on an APU, whose firmware carve-out alone is not the usable pool); `statsKey` = `amd:<pci address>`, so this composes with #43 if both land.
- `stats_amd_linux.go`: each tick samples `mem_info_*_used` (from the same side of the APU/discrete rule as the capacity, so used never exceeds total) and `gpu_busy_percent`; a missing or out-of-range value is dropped rather than published as `0`.
- `gpu_linux.go` / `stats_linux.go`: AMD rows are appended after the NVIDIA rows; `telemetryValid` turns true once a sample lands.
- README section: what is read, the unified-pool rule, the join key, ceilings (`amdgpu` only; busy % is the driver's coarse figure). Temperature/power are deliberately left out of this PR — they are not part of the current wire contract.

## How tested
- `go vet`, `go test ./...` on a fake sysfs tree (naming, APU vs discrete capacity, connector/`renderD` aliasing, the join by PCI address, sampling), run natively for the Linux build on an arm64 Linux host.
- Live on a Ryzen 5 5625U (Barcelo) node: name, pool, used bytes and busy % all present; `telemetryValid` true.

## Risk
Low: Linux-only, additive rows; a host without an AMD card costs one `ReadDir` per tick.

Draft, like #90 — opened from the fork for maintainers to shape.


sherief-nv:
This seems to be a dupe of #43 which we plan on merging very soon.

